PR #11, opened 2016-08-23
Added footer message for cf-component-card/CardContent

thellimist:
**This is the result**
![alt text][image1]

I added footer message because it was missing in cf-ui and was present in cloudflare.com

**Example of footer messages in cloudflare.com**
![alt text][image2]

[image1]: http://i.imgur.com/DpY9fzl.png 
[image2]: http://i.imgur.com/3o1fqa5.png

### DIFF
--- a/packages/cf-component-copyable-textarea/src/CopyableTextarea.js
+++ b/packages/cf-component-copyable-textarea/src/CopyableTextarea.js
@@ -38,7 +38,7 @@ class CopyableTextarea extends React.Component {
     let success;
     try {
       success = document.execCommand('copy');
-    } catch(e) {
+    } catch(exception) {
       success = false;
     }
 
--- a/packages/cf-component-copyable-textarea/src/CopyableTextarea.js
+++ b/packages/cf-component-copyable-textarea/src/CopyableTextarea.js
@@ -38,7 +38,7 @@ class CopyableTextarea extends React.Component {
     let success;
     try {
       success = document.execCommand('copy');
-    } catch(exception) {
+    } catch(err) {
       success = false;
     }
 
--- a/packages/cf-util-text/src/text.js
+++ b/packages/cf-util-text/src/text.js
@@ -4,9 +4,9 @@ function getText(id) {
   if (!text[id]) {
     console.error('Missing text for: "' + id + '"');
     return id;
-  } else {
-    return text[id];
-  }
+  } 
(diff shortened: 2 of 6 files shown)


jwineman:
@wyuenho @marksteyn @terinjokes @toekneestuck

wyuenho:
LGTM

jwineman:
What about cf-component-text for the the footer text?
https://cloudflare.github.io/cf-ui/#cf-component-text
```
<Text type="muted">Hello World</Text>
```

Do we care if our components have dependencies on each other?
Are there any examples in the app where we don't use this space for displaying a timestamp of the last time the card was modified?  If so is it worth renaming this to something more specific to its function?  Ex timestamp?

wyuenho:
Calling it footer or foot-note is fine. I like using `<Text>` better.

screen shot 2016-08-23 at 18 21 14

packages/cf-util-text/src/text.js
```diff
@@ -4,9 +4,9 @@ function getText(id) {
   if (!text[id]) {
     console.error('Missing text for: "' + id + '"');
     return id;
-  } else {
-    return text[id];
-  }
+  } 
+  
+  return text[id];
```
jwineman: Why did this change?

wyuenho: Linter probably

packages/cf-component-copyable-textarea/src/CopyableTextarea.js
```diff
@@ -38,7 +38,7 @@ class CopyableTextarea extends React.Component {
     let success;
     try {
       success = document.execCommand('copy');
-    } catch(e) {
+    } catch(exception) {
```
wyuenho: `e` is the convention, there are no exceptions in JS, just errors.

thellimist:
@wyuenho Edited

jwineman:
Still not using `<Text>`

wyuenho:
@thellimist say he couldn't get this package depend on cf-component-text

jwineman:
Why not?

thejameskyle:
Adding a cross-dependency: 

1. Add the `"name": "^version"` to `package.json`
2. Run `npm run bootstrap` again

thejameskyle:
*vanishes into the night*

PR #14, opened 2016-08-24
Modified CardContent/footerMessage to use <Text>

thellimist:


example/styles.css
```diff
@@ -160,9 +160,6 @@ button:not(.cf-btn) {
 }
 
 .cf-card__footer_message {
```
jwineman: We should be able to delete all the custom css you added right?  `<Text/>` css should already be handled from when the examples for that component were created.

thellimist: I can not remove `margin-top: 1.5em;`

jwineman: Why? `.cf-card__footer_message` shouldn't even be used anymore since `<Text/>` relies on a different set of CSS classes right?

wyuenho: `.cf-card__content div[class^="cf-text"] {
    margin-top: 1.5em
}`

packages/cf-component-card/src/CardContent.js
```diff
@@ -16,7 +17,7 @@ class CardContent extends React.Component {
         {this.props.children}
         { this.props.footerMessage ? (
           <div className="cf-card__footer_message">
```
wyuenho: No need for this wrapper

jwineman: We can remove `className` though right?  If not I think we need to rename it because its not following the BEM naming structure.

example/styles.css
jwineman: Make sure you make this change to the internal styles in WWW-Style as well @thellimist.

packages/cf-component-card/src/CardContent.js
wyuenho: Just use <Text> directly, no need to wrap it just so we can apply a CSS class with 1 property.

jwineman: Ah I misread your first comment as saying we _did_ need it.  Yeah, I agree, just remove the wrapper.

example/styles.css
thellimist: @wyuenho Should we need to consider if someone else using `<Text>` inside `<CardContent>`?

Example:
```
<CardContent title="xxxx" footerMessage="xxx">
       <Text case="titlecase">hello world</Text>
</CardContent>
// or
<CardContent title="xxxx" footerMessage="xxx">
       <Text type="error">Hello World</Text> 
</CardContent>
```

wyuenho: .cf-card__content > div[class^="cf-text"]:last-child {
}

thellimist: Without footer message
``` 
<CardContent title="xxxx">
       <Text type="error">Hello World</Text> 
</CardContent>
```

wyuenho: Good point. Well, I guess this PR is not-needed now :)

wyuenho:
@thellimist Raised a good point, if we use `<Text>`, there's no way to distinguish the last child from the foot text when CSS is applied. 

Do you think we should make cf-component-text to accept arbitrary additional class names or just close this PR and leve cf-component-layout as is?

jwineman:
I don't see a problem with wrapping it if we need to differentiate it in the styles, it just needs a better name in that case - `.cf-card__content-footer` probably works.

wyuenho:
We should try to keep the DOM elements we create to a minimal. We shouldn't have to create extra div just to apply a style. What we have now is enough.

jwineman:
I'm okay with closing the PR as well if you think thats a better approach.

jwineman:
Do we need to revert https://github.com/cloudflare/cf-ui/pull/11 too?

wyuenho:
No, #11 is exactly what should happen.

PR #15, opened 2016-08-31
PI-764: Updating cf-component-loading to depend on cf-component-icon …

jwineman:
Externally `cf-component-loading` is styled fine.  

Internally there are no styles for `cf-component-loading` but `cf-component-icon` does contain `cf-component-loading` styles.  

I updated the component to depend on `cf-component-icon` styles because I think `<Loading\>` is more explicit than `<Icon label="border" type="gear" loading/>`.

### DIFF
--- a/packages/cf-component-loading/src/Loading.js
+++ b/packages/cf-component-loading/src/Loading.js
@@ -1,11 +1,10 @@
 const React = require('react');
+const Icon = require('cf-component-icon');
 
 class Loading extends React.Component {
   render() {
     return (
-      <div className="cf-icon--loading" role="status">
-        <span className="cf-icon--loading__span">Loading</span>
-      </div>
+      <Icon type="loading" label="Loading"/>
     );
   }
 }
--- a/packages/cf-component-loading/src/Loading.js
+++ b/packages/cf-component-loading/src/Loading.js
@@ -3,8 +3,8 @@ const React = require('react');
 class Loading extends React.Component {
   render() {
     return (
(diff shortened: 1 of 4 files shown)


jwineman:
@toekneestuck @wyuenho @marksteyn @terinjokes @thellimist

packages/cf-component-loading/src/Loading.js
```diff
@@ -3,8 +3,8 @@ const React = require('react');
 class Loading extends React.Component {
   render() {
     return (
-      <div className="cf-loading" role="status">
-        <span className="cf-loading__label">Loading</span>
+      <div className="cf-icon--loading" role="status">
```
thejameskyle: Modify and use `<Icon>` for this, the classnames are not meant to be part of the public interface. They should be contained solely within the package that creates them.

jwineman: So delete `cf-component-loading`

thejameskyle: I would keep `<Loading/>` as well because it should abstract away the a11y details. Just depend on icon

packages/cf-component-loading/src/Loading.js
```diff
@@ -1,11 +1,10 @@
 const React = require('react');
+const Icon = require('cf-component-icon');
 
 class Loading extends React.Component {
   render() {
     return (
-      <div className="cf-loading" role="status">
```
thejameskyle: You should have the role too

PR #16, opened 2016-09-08
PI-764: Adding back role to cf-component-loading to be consistent wit…

jwineman:


jwineman:
@terinjokes @toekneestuck @wyuenho @marksteyn @thellimist

wyuenho:
LGTM

packages/cf-component-icon/test/Icon.js
```diff
@@ -40,4 +40,12 @@ describe('Icon', function() {
       <i className="cf-icon cf-icon--info-sign"></i>
     );
   });
+
+    it('should render role', function() {
+    assertEqualJSX(
```
thejameskyle: What even is indentation

jwineman: lol I'll fix.  Had to edit in sublime cause intellij kept crashing for some reason.  Need you to add me as a publisher to all the packages though.

terinjokes: @thejameskyle indentation adds so many wasted bytes to a git clone. if we remove all indentation cloning will be much faster.

PR #25, opened 2016-11-28
remove cf preset and upgrade eslint to 3

wyuenho:
* all the class properties have been rewritten to use boring es2015 syntax
* eslint 3
* new eslint config for all packages
* cleaned up all lint errors
* cleaned up most of the `findDOMNode` usage except Tooltip


packages/cf-builder-card/src/CardBuilder.js
```diff
@@ -5,30 +5,14 @@ const {
...
   CardSection,
   CardContent,
-  CardMessages,
   CardControl,
   CardDrawers,
   CardPropTypes
 } = require('cf-component-card');
-const DynamicContent = require('cf-component-dynamic-content');
-const markdown = require('cf-util-markdown');
+const DynamicContent = require('cf-component-dynamic-content');const markdown = require('cf-util-markdown');
```
nkcmr: @wyuenho any reason the newline was taken out?

wyuenho:
This should be ready to go. Will merge tomorrow.

PR #30, opened 2016-12-08
Fix import in example

koddsson:
Seems to me that we should have the npm package name in the require rather than the relative path :P


thellimist:
I agree it's a hassle to convert those when trying copy from the docs.

FYI The website is auto generated https://github.com/cloudflare/cf-ui/blob/master/gulpfile.babel.js#L499

PR #39, opened 2016-12-20
Adds a collapsable panel component

jculvey:
Not really all that fancy. You click the link, it shows you more stuff.

Screencap of example:

image

### DIFF
--- a/packages/cf-component-collapsable-panel/.npmignore
+++ b/packages/cf-component-collapsable-panel/.npmignore
@@ -1,0 +1,6 @@
+node_modules
+*.log
+src
+coverage
+test
+example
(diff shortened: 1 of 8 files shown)


koddsson:
I have something similar where I encapsulated [react-collapse](https://github.com/nkbt/react-collapse) to get that sweet animation. Would you consider something like that in this component?

jculvey:
@koddsson yeah totally. Is it ready to be used? I can toss this and we can use it instead.

jculvey:
Actually, after looking at react-collapse, I think I can make a single component similar to the [expansion panel](https://material.io/guidelines/components/expansion-panels.html#expansion-panels-behavior) from material design that will work for both use cases like this and in list form for something else I need.

PR #40, opened 2016-12-21
Implements a new multi step modal wizard component

jculvey:


packages/cf-component-wizard/README.md
```diff
@@ -0,0 +1,99 @@
...
+  GatewayProvider
+} = require('react-gateway');
+
+class Component extends React.Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeStep: 'step1',
```
koddsson: Will this ever be something other than a numerical value? Should we have this just as a integer so that something like this would be possible?

```js
this.setState((state) => {
  return {activeStep: state.activeStep + 1}
});
```

packages/cf-component-wizard/README.md
```diff
@@ -0,0 +1,99 @@
...
+      isOpen: false
+    };
+
+    this.handleRequestOpen = this.handleRequestOpen.bind(this);
+    this.handleRequestClose = this.handleRequestClose.bind(this);
```
koddsson: Should we just get rid of this and do the binds where the function is used? Either that or change line 58 to do the same as this.

Only reason why I'm suggesting it is because I'm more use to the pattern of binding before passing down.

packages/cf-component-wizard/example/basic/component.js
```diff
@@ -0,0 +1,96 @@
...
+const {render} = require('react-dom');
+const Wizard = require('../../src/index');
+const {Button} = require('cf-component-button');
+
+const {
+  GatewayDest,
+  GatewayProvider
+} = require('react-gateway');
+
+class Component extends React.Component {
```
koddsson: I could be wrong but shouldn't we import the component and use it in examples instead of re-implementing?

packages/cf-component-wizard/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "cf-component-wizard",
+  "description": "CloudFlare Wizard Component",
+  "version": "4.0.0",
+  "main": "lib/index.js",
+  "author": "James Kyle <jkyle@cloudflare.com>",
```
koddsson: s/Kyle/Culveyhouse/

example/styles.css
```diff
@@ -1102,6 +1102,20 @@ a.cf-link--disabled {
 }
 
 /**
+ * Wizard
+ */
+
+.cf-wizard .cf-modal__footer {
+  display: flex;
```
jwineman: Does the component use flexbox or just the example?  If the component requires flexbox can we use `cf-component-flex` instead?

packages/cf-component-wizard/src/Wizard.js
```diff
@@ -0,0 +1,115 @@
...
+    this.handleNextClick = this.handleNextClick.bind(this);
+    this.handleBackClick = this.handleBackClick.bind(this);
+  }
+
+  // The Progress component expectes a similar set of steps
+  // with a few less fields
+  extractProgressSteps(wizardSteps) {
+    return wizardSteps.map((step) => {
+      const {id, label, disabled} = step;
+      return {id, label, disabled};
```
koddsson: Why do we need this?

packages/cf-component-wizard/example/basic/component.js
koddsson: Oh, I'm mistaking very much. Ignore me!

packages/cf-component-wizard/src/Wizard.js
```diff
@@ -0,0 +1,115 @@
...
+              <Progress active={activeStep}
+                onChange={onStepChange}
+                steps={this.extractProgressSteps(steps)} />
+              <div className="cf-wizard__step-content">
+                {step.content}
+              </div>
+            </div>
+          </ModalBody>
+          <ModalFooter>
+            {!isFirstStep && (
```
koddsson: Should this be in in the `<ModalActions>` with the others?

packages/cf-component-wizard/README.md
```diff
@@ -0,0 +1,99 @@
...
+              id: 'step1',
+              label: 'Step 1',
+              content: (<div>Step 1</div>),
+              disabled: true
+            },
+            {
+              id: 'step2',
+              label: 'Step 2',
+              content: (<div>Step 2</div>),
+              disabled: false
```
koddsson: What does this prop do?

wyuenho:
I'm not sure this should live in cf-ui now. I think for me, things that live in cf-ui should:

1. generic enough to take care of all of our use cases and
2. also useful for everyone else outside of Cloudflare and

This doesn't seem to satisfy 1) and probably 2). 

A. The ProgressBar shouldn't be required, and it isn't for our existing Backbone multi-step modal.
B. We have many modals that don't have ModalHeader.
C. We often have modals that dynamically inject or remove steps depending on the result of some previous steps.
D. Many modals have completely different arrangement of buttons. The number could be different, text, color and behavior can all be completely different.

C and D is the biggest pain point for me for our existing Backbone multi-step modal. It's not clear to me how to use a bunch of static configuration to achieve C and D. I'm inclined to think the abstraction we generalize should end at `cf-component-Modal`. If you have a good idea of how to deal with these constraints, please feel free. Otherwise, I prefer to decline.

example/styles.css
jculvey: It realy need flexbox to render the buttons properly anchored to the left and right.

`cf-component-flex` won't quite do it today. It doesn't support much, just 2 spacing options. I kinda think we should add justifyContent, alignItems, and alignContent support to it, which would solve this problem and many others.

packages/cf-component-wizard/src/Wizard.js
jculvey: spacing issues....

packages/cf-component-wizard/src/Wizard.js
jculvey: Technically, we kinda don't. I purposely made the steps data structure match between wizard and progress, except that the wizard's steps has more things. I figured people might dislike variables leaking into progress, so I added this.

Issue #33, opened 2016-12-19
Allow all UI components to accept extra class names and modifiers


wyuenho:
A lot of the components have hard coded class names and states that get appended to the class names, but we can't append to the class names when customizing, either to override the styles in a specific component in a certain context, or introduce contextual BEM modifiers that will only be apparent at the use site.

jwineman:
How is this different from https://github.com/cloudflare/cf-ui/pull/31?

Not exposing extra class names is by design.

wyuenho:
The rationale [here](https://github.com/cloudflare/cf-ui/blob/master/discussions/no-customizing-classnames.md) doesn't seem to make sense anymore. In practice, CSS styles will leaks into every component because inheritance, and there's no way to firewall that from the JS side. Designers also have a lot of requirements that we have to implement as exceptions to the rules, we need a way to override the default classes on demand.

jwineman:
I think a lot of these issues stem from the fact that we haven't finished the backbone migration yet. 

The component based styles don't use inheritance, the BEM styles are specific to the component and don't inherit styles from anywhere else.

I would argue we want to push back against designers making exceptions to the rules - every exception want to create makes the code base harder to maintain.

wyuenho:
Typographic styles are inherited almost all the time, they affect sizes, in which there's no one-size-fits-all approach will work.

The major problem now is there's no way to pass the props down to the inner components. Using page level styles is a non-solution. Many components like modal, tooltip etc don't get attached to a specific page. The form elements in them also often have style adjustments that we have to deal with. Without being able to add specific class names and modifiers, it's very hard to impossible to target these elements.

jwineman:
`<Heading/>` and `<Text/>` and `cf-component-typography` don't have inherited styles right?

I think `<Modal/>` is a special case that might warrant exception.  What about exposing an CSS ID on `<Modal/>` instead of full on style/class props?

What is an example of a form element that can't be styled at the page level?

wyuenho:
We have global typographic styles, and we'll always have them, unless you introduce a `<Body />` component. IMHO, those typographic cf-ui elements are superfluous. We aren't going to use `<Heading/>` in every place we have to use a heading....

wyuenho:
ID is also a terrible idea. Nobody is going to check if that ID exists already before putting an ID

jwineman:
I think the point of using a UI library is to _avoid_ global styles and I disagree we'll always need to have them.  

[...]

PR #43, opened 2016-12-21
Optional custom class names for table components

wyuenho:
This is a part of #33 

Sometimes, we'll need to have CSS classes that target the rows and columns. Since technically the HTML spec doesn't allow us to wrap elements around `<tr>`,`<th>` and `<td>`, we have to use custom class names to avoid having a mess of CSS selectors all over the place.

This PR fixes this issue.


thejameskyle:
For the record, it was intentional not to have `classNames` on any of this components. It's an anti-pattern.

PR #44, opened 2016-12-22
Fix issue with not being able to focus and select the textarea

wyuenho:
React 15 upgrade broke CopyableTextarea and I didn't catch this because there were no tests. The ref returned was the Component instead of the wrapped DOM element, which has `focus()` and `select()`.

This rewrote CopyableTextarea to be more straightforward. An `onClick` prop has been added to Textarea as well to enable copying the programmatically selected text from CopyableTextarea.

You can also supply the help text from the props, just like what the README says.

### DIFF
--- a/packages/cf-component-copyable-textarea/package.json
+++ b/packages/cf-component-copyable-textarea/package.json
@@ -10,9 +10,11 @@
   },
   "dependencies": {
     "cf-component-textarea": "^2.0.0",
-    "cf-util-text": "^2.0.0",
     "react": "^0.14.2 || ^15.0.0-0",
     "react-dom": "^0.14.2 || ^15.0.0-0"
   },
-  "devDependencies": {}
+  "devDependencies": {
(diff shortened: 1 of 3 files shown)


hturan:
I don't have approval rights, but I approve 😄

My only question is more of a repo-wide philosophical one regarding `cf-util-text`. It seems a bit opaque/user-hostile to `console.error` if a user doesn't know to call `setText` on an arbitrary string before rendering a component. Are we going to continue using `cf-util-text` for future components and focus on documentation? Or is the future direction going to be passing in localised copy as props?

wyuenho:
I think we should deprecate cf-util-text. First of all, there are many gettext-like translation libraries out there, cf-util-text is nowhere close to their functionality. I don't want to grow yet another translation package that's 1/10 as useful as others. Secondly, any sensible UI library should allow the user to change some text in the most straight forward way. cf-util-text simply isn't needed.

PR #47, opened 2016-12-27
Add card blocks and wide card modifier

prestonp:


packages/cf-component-card/src/CardControl.js
```diff
@@ -3,16 +3,19 @@ const {PropTypes} = React;
 
 class CardControl extends React.Component {
   render() {
+    const className = 'cf-card__control' + (this.props.wide ? ' cf-card__control--wide' : '');
```
jculvey: I've found the [the `classnames` library](https://github.com/JedWatson/classnames) library to be useful at handling this particular oft-occurring problem.

prestonp: Hey that's really cool! I think this should be addressed across the entire repo so I opened  issue #49

jculvey: Awesome!

wyuenho:
Why do we need yet another extra card block as opposed to just a div?

prestonp:
I just wanted to make it an easy component. Of course, anything customization could just be defined downstream but we use this pattern a lot for multiple sections within a CardControl

wyuenho:
Don't we already have a bunch of layout components you can use to align the blocks?

prestonp:
The closest thing I saw was .cf-form__fieldset_content from the Form components but would still take some modification. I think this is intuitive and easy to read:

```jsx
<CardControl>
  <CardBlock><button ... /></CardBlock>
  <CardBlock><button ... /></CardBlock>
  <CardBlock><button ... /></CardBlock>
</CardControl>
```

Issue #22, opened 2016-11-28
Remove experimental draft ES constructs


wyuenho:
We are using a lot of experimental stuff like class properties, a withdrawn syntax for `export default` and object spread. Since only object spread seems to be going somewhere, I propose we remove class-properties and export-extensions. This will also stop ESLint from complaining without having to do weird things to it.

thejameskyle:
Class properties is a pretty safe proposal at this point, it's stage 2 and the only debate is over if `=` should be changed to something like `:=` which we'll probably discuss at this week's tc39 meeting.

Also, dropping `babel-preset-cf` is definitely not what you should be doing. Keep control over your own preset and version it yourself.

wyuenho:
@terinjokes said static properties should be `{configurable: false, enumerable: true, writable: true}` but babel currently sets them all to true. I'd really like to minimize semantic differences. There's enough of that in the ES2015 preset against the real spec as is. We'll discuss more about class properties internally, decisions haven't been made yet.

thejameskyle:
Then turn on `spec` mode for https://github.com/cloudflare/babel-preset-cf/blob/master/index.js#L7 if you're that concerned about it.

```diff
-     require("babel-plugin-transform-class-properties"),
+     [require("babel-plugin-transform-class-properties"), { spec: true }],
```

wyuenho:
I just read the TC39 minutes and the issues in the repo. It seems like there are still talks about what `this`, `super`, `arguments` should be bound to, interactions with decorators, private fields and inheritance. There's also computed properties to consider. There's too much still up in the air now.

Separately, `babel-present-cf` isn't being used much both external and internally. There are a number of problems with this preset. The `loose` options being one, incomplete ES2015 syntax and experimental syntax being another. The current preset makes it easy to introduce stable syntax but difficult to convert unstable or non-spec-compliant syntax to spec. It's far easier to remove dependents on it now then later. We can come up with an updated company-side application preset and library preset later if we desire.

thejameskyle:
Silly

wyuenho:
:)

wyuenho:
Fixed in PR#25

PR #53, opened 2017-01-11
Init flow

thejameskyle:
I had this branch started forever ago. I finished it off because I want to show off Flow with component frameworks like this.


thejameskyle:
Note: This doesn't do anything right now to distribute types once published. Didn't bother setting that up

thejameskyle:
Also all the `propTypes` are left in place because that's actual work and I know you aren't using Flow yet.

.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": ["es2015", "react"],
-  "plugins": ["transform-object-rest-spread"]
+  "presets": ["cf"],
```
wyuenho: Noooooooo we don't want to maintain cf preset anymore before an internal style guide for ES6 emerge

packages/cf-builder-card/src/CardBuilder.js
```diff
@@ -48,21 +66,7 @@ class CardBuilder extends React.Component {
   }
 }
 
-CardBuilder.propTypes = {
```
wyuenho: I'm guessing Flow has trouble type checking props if we attach prop types on the class like this?

packages/cf-builder-card/src/cardReducer.js
```diff
@@ -1,10 +1,12 @@
-const ActionTypes = require('./CardActionTypes');
+// @flow
 
-const initialState = {};
+import type {CardAction, CardState} from './CardTypes';
```
wyuenho: Yay imports!

thejameskyle: These aren't normal imports, they just get stripped out. There isn't another way to import types in Flow.

packages/cf-builder-card/src/CardBuilder.js
thejameskyle: You'd have to re-specify the types for `propTypes` `defaultProps` and all the `contextProps` stuff. Otherwise Flow wouldn't consider these actual members of the class.

.babelrc
thejameskyle: Well then you should re-specify all the config inside the existing one: https://github.com/cloudflare/babel-preset-cf/blob/master/index.js#L2-L24

There's a good reason for every single line in that file. I offered to go over it all when I was still at cf and no one took me up on that.

wyuenho: https://github.com/cloudflare/cf-ui/issues/22

We can't even use cf preset in our dashboard. It's a very strange collection of extremely experimental constructs and a very constraining, but not spec-conforming subset of ES2015. It's essentially giving us a different language that's neither a superset nor subset of ES2015. It's too complicated. Also, we want to use ES6 modules and Webpack at some point so consumers of cf-ui can tree shake the generated library code.

packages/cf-builder-card/src/cardReducer.js
wyuenho: No I'm not saying it's wrong, I'm actually fine with this. In fact I'd like to use normal imports anyway.

packages/cf-builder-card/src/CardBuilder.js
wyuenho: That's fine, we'd have to change the types anyway to take full advantage of Flow (i.e. functions). I'm just trying to understand if Flow requires us to use class properties.

.babelrc
thejameskyle: I already knew what you thought. But whatever

packages/cf-builder-card/src/CardBuilder.js
thejameskyle: Right now the alternative is either:

```js
type Props = {
  propName: string;
};

type DefaultProps = {
  propName: string
};

type State = {
  stateName: string
};

class MyComponent extends React.Component<DefaultProps, Props, State> {
  constructor(props, context) {
    super(props, context);
    this.state = {
      stateName: this.props.propName;
    };
  }
  render() {
    return <div/>
  }
}

MyComponent.propTypes = {
  propName: React.PropTypes.string.isRequired
};

MyComponent.defaultProps = {
  propName: 'foo'
};
```

Or:

```js
class MyComponent extends React.Component<*, *, *> {
  // ...
}

MyComponent.propTypes = {
  propName: React.PropTypes.string.isRequired
};

MyComponent.defaultProps = {
  propName: 'foo'
};
```

> ^ Please do not do this second option, `*` is magic that does not work the way you expect it to.

In the future when you are using flow everywhere you can switch to this:

```js
type Props = {
  propName: string;
};

class MyComponent extends React.Component {
  props: Props;
  
  static defaultProps = {
    propName: 'foo'
  };

  state = {
    stateName: this.props.propName;
  };

  render() {
    return <div>
  }
}
```

packages/cf-builder-card/src/cardReducer.js
thejameskyle: You know it's weird to be concerned about class properties while being okay with imports right? Imports are less standardized than class properties.

wyuenho: ES6 modules are standardized, how to load them is not since there are 3 (4?) related proposals floating around. But it doesn't matter since we don't use any of the proposed loading mechanism.

packages/cf-builder-card/src/CardBuilder.js
wyuenho: Option 1 looks good, since it's easy to add to existing code.

thejameskyle: Option 1 might _look_ good, but it is still not good. It's much better for Flow to know definitively what properties are allowed so that variance works as expected.

packages/cf-builder-card/src/cardReducer.js
thejameskyle: That would be fine and dandy except the TC39 is currently trying to figure out what to do with modules because they can't change some semantics about them because Babel users are already using them.

So no, there is no "standardized" module spec. There's just a syntax which is now causing problems and we might need to break things for everyone.

But if I'm wrong, then I'd love to hear the context you have that I don't.

packages/cf-component-checkbox/src/Checkbox.js
```diff
@@ -1,10 +1,27 @@
+// @flow
+
 const React = require('react');
 const {PropTypes} = React;
 
+type InputElementEvent = Event & {
+  target: HTMLInputElement
```
thejameskyle: Should replace all of these with `currentTarget` since this is unsound due to bubbling.

packages/cf-builder-card/src/cardReducer.js
wyuenho: The ES2015/6 module semantics are clearly defined in chapter 8, 9, and 15 in the ECMAScript 2015/6 spec.

AFAIK from November TC39 minutes, they mainly talked about [node module interops with ESM](https://github.com/nodejs/node-eps/blob/master/002-es6-modules.md), dynamic imports and whether `export * from` should export a default as well.

For node/ESM interops, the things at stake are 1) how to parse the source unambiguously to distinguish whether it's a script or a module, 2) deal with unbound names in circular import. Both of which are stage 1. I don't want to go into details but AFAIK, 1) has [an answer](https://github.com/nodejs/node-eps/blob/5dae5a537c2d56fbaf23aaf2ae9da15e74474021/002-es6-modules.md#51-determining-if-source-is-an-es-module) in Node, they are just trying to find out if it would be useful for the Web as well. 2) also has 3 options, no breaking change was proposed.

Webpack (as of rc3, rc4 broke this) seems to follow the Node EPS closely with some necessary small differences. We would love to enable downstream users to tree-shake cf-ui. We can only do that with import/export so Webpack can compile into a form that it can tree-shake. Not that we have much to tree-shake, but it'd be nice.

There's no plan to change module semantics for dynamic imports, Chrome is implementing it now. We are currently using Babel's dynamic import syntax plugin to pass `import()` thru to Webpack.

The consensus for exporting the default in `export * from` is there isn't consensus, so we are still in the steady state that's ES2015/6 module semantics, and the `export default from` and `export * as ns from` stage 1 proposals. I assume you are talking about this when you said "break things", which it might if it were adopted.

If you are talking about a Babel module transform bug, I can tell you for sure we are not going to transform ES6 module with Babel. `{module: false}` will most certainly be in `.babelrc`.

If you are talking about ES6 modules not implemented in browsers, that's because they are all waiting for the dynamic imports/loader spec, in which the unambiguous grammar issue kinda a big deal.

wyuenho: I think the takeaway is, when everything is bundled, the modules will be ordered and eval'ed statically except things loaded by `import()`, in which Webpack will wrap into an "installable" static order anyway. We don't effect any side-effect in module exports that can't be replicated statically, and we certainly don't have any circular imports. There doesn't seem to be any proposed changes that will affect us.

tajo:
I want to close this for now. Flow is on our long-term todo list but CSS in JS transition is more important and it's going to make this PR even less mergeable than it is now.

@jwineman ?

jwineman:
I have a branch on Partners where I copied a lot of James' config.  We can always reference this branch when we implement Flow.  I have a decent understanding of Flow now.  I'm fine closing.

Issue #46, opened 2016-12-22
All pure components should accept all DOM props


wyuenho:
cf-ui has a ton of pure components that only do 2 things:

1. Apply BEM classes
2. Prevents outside code from directly setting any event handlers and props on the contained React DOM element that's not whitelisted by the pure components themselves.

2) is preventing us from setting event handlers on form elements and sometimes completely valid HTML 5 properties. All pure components should allow all of the attributes and event handlers accepted by the wrapped elements themselves when possible. Special handling for some attributes and event handlers might be needed, but they should be dealt with on a component-by-component basis and should not hinder this issue.

koddsson:
Oh this has been open for a while and nobody has any comments yet! Let me take take a stab at it :)

What's the downside of doing something like:

```js
<input {...props} />
```

in our components?

I can see the benefit being for example we don't have to make a PR to cf-ui to make sure that we can [disable a toggle](https://github.com/cloudflare/cf-ui/pull/55) which would speed up our process.

wyuenho:
Sure. Let's ID a list of components we want to do this first. As it stands, a lot of the typographic components are not useful at all and I don't see the value of working on them further.

wyuenho:
Also, we might not want all props to be expanded, just those we know are accepted by HTML5.

koddsson:
Ok, we could keep a list maybe of allowed props? What would be the downside of expanding all props?

wyuenho:
I have no good reason to continue with a whitelist approach so far, just something we need to keep in mind. Best go thru all the components now and see if there are idioms in them that have a high probability in resulting accidental overwrites of attributes.

jwineman:
The only downside I can think of is `propTypes` don't warn when you [pass in extra props](https://github.com/facebook/react/issues/1587).  I can't think of any scenarios where this would actually affect us negatively though.

wyuenho:
This is a living ticket as PRs that allow arbitrary props to be passed into the cf-ui components are opened and merged, until all components are fixed.

koddsson:
@wyuenho In order for this to be actively worked on (and maybe one day finished) could you list up the places where we are not passing the props through so we have a idea of how much is left of this ticket?

Maybe it even goes hand-in-hand with the fela migration since `createComponent` accepts a `passThroughProps` parameter?

tajo:
[...]

PR #55, opened 2017-01-11
Support disabled toggles

prestonp:
toggle-disabled


wyuenho:
Gosh we really need to deal with #46 so we can stop these kinds of PRs

PR #58, opened 2017-01-13
Pass props from cf-component-input to the inner input

koddsson:
See #46. I'm not quite sure everyone is gonna be ok with this but I figured this could be a way to get more people involved.

### DIFF
--- a/packages/cf-component-input/src/Input.js
+++ b/packages/cf-component-input/src/Input.js
@@ -19,14 +19,7 @@ class Input extends React.Component {
     }
 
     return (
-      <input
-        type={this.props.type}
-        className={className}
-        name={this.props.name}
-        value={this.props.value}
-        onChange={this.handleChange}
-        placeholder={this.props.placeholder}
-        autoComplete={this.props.autoComplete}/>
+      <input className={className} onChange={this.handleChange} {...this.props} />
     );
   }
 }
--- a/packages/cf-component-input/test/Input.js
+++ b/packages/cf-component-input/test/Input.js
@@ -88,4 +88,22 @@ describe('Input', function() {
     );
   });
 
+  it('should pass all props down to the inner input', function() {
+    assertEqualJSX(
+      <Input
+        disabled
+        type="number"
+        data-shadowed={true}
+        min={3}
+      />,
+      <input
+        className="cf-input cf-input--number"
+        disabled="true"
+        type="number"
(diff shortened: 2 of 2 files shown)


packages/cf-component-input/src/Input.js
```diff
@@ -19,14 +19,7 @@ class Input extends React.Component {
...
     return (
-      <input
-        type={this.props.type}
-        className={className}
-        name={this.props.name}
-        value={this.props.value}
-        onChange={this.handleChange}
-        placeholder={this.props.placeholder}
-        autoComplete={this.props.autoComplete}/>
+      <input className={className} onChange={this.handleChange} {...this.props} />
```
wyuenho: `className` needs to be handled separately. I think we want to join the passed in one with the default one.

prestonp: Don't think this will use `this.handleChange` as expected

>The specification order is important. Later attributes override previous ones.
https://gist.github.com/sebmarkbage/07bbe37bc42b6d4aef81#spread-attributes

prestonp: I ran the demo page as a testing playground and this breaks the input. 

This seemed to fix
```jsx
  render() {
    let { className, ...props } = this.props;

    className = className + ' cf-input cf-input--' + props.type;

    if (props.invalid) {
      className += ' cf-input--invalid';
    }

    return (
      <input {...this.props} className={className} onChange={this.handleChange}  />
    );
  }
```

koddsson: Good catch! I'll add test for this scenario to this PR

wyuenho: @prestonp The handleChange issue is what I was alluring in the ticket. We need to deal with existing whitelisted props on a case by case basis. I think if an event handler is supplied externally in the props, it should mean that the user wants the component to use that one instead of the internal one.

Alternatively, the internal event handler should call the externally supplied event handler. However, **when** (before? after? case by case?) we should call it is not clear. Back in the days when `addEventListener` was available, the order is attachment order, and the convention is to write the event handlers in a way that do not interfere with or depend on each other.

Another option is open up the internal event handler for external event handlers to call, and if an event handling prop is supplied, use it and assume that it will call the internal event handler as it see fit. My preference is this option.

wyuenho:
I've merge this into PR #60, closing this one.

PR #60, opened 2017-01-16
Pass all props to input

wyuenho:
Part of #46 . Supersedes #58

### DIFF
--- a/packages/cf-component-input/src/Input.js
+++ b/packages/cf-component-input/src/Input.js
@@ -2,9 +2,6 @@ const React = require('react');
 const {PropTypes} = React;
 
 class Input extends React.Component {
-  constructor(props) {
-    super(props);
-  }
 
   render() {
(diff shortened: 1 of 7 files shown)


packages/cf-component-input/src/Input.js
```diff
@@ -4,29 +4,23 @@ const {PropTypes} = React;
 class Input extends React.Component {
   constructor(props) {
```
hturan: Do we still need this empty `constructor`?

packages/cf-component-input/src/Input.js
```diff
@@ -4,29 +4,23 @@ const {PropTypes} = React;
 class Input extends React.Component {
```
hturan: Could this now be a stateless component?

wyuenho: Internal discussion: can't have refs on functional component

PR #63, opened 2017-01-17
[TextArea] Pass any prop down to inner component

koddsson:
Replaces #61 and #62

### DIFF
--- a/packages/cf-component-textarea/src/Textarea.js
+++ b/packages/cf-component-textarea/src/Textarea.js
@@ -16,14 +16,13 @@ class Textarea extends React.Component {
   }
 
   render() {
-    let {className, ...props} = this.props;
-    
-    className = 'cf-textarea';
+    const {className, ...props} = this.props;
+    const classes = ['cf-textarea', className];
 
-    if (this.props.disabled) className += ' cf-textarea--disabled';
-    if (this.props.readOnly) className += ' cf-textarea--readonly';
+    if (this.props.disabled) classes.push('cf-textarea--disabled');
+    if (this.props.readOnly) classes.push('cf-textarea--readonly');
 
-    return <textarea className={className} {...props} />;
+    return <textarea className={classes.join(' ')} {...props} />;
   }
 }
 
--- a/packages/cf-component-textarea/src/Textarea.js
+++ b/packages/cf-component-textarea/src/Textarea.js
@@ -16,22 +16,14 @@ class Textarea extends React.Component {
   }
 
   render() {
-    let className = 'cf-textarea';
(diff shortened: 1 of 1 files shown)


packages/cf-component-textarea/src/Textarea.js
```diff
@@ -16,22 +16,14 @@ class Textarea extends React.Component {
...
-        className={className}
-        name={this.props.name}
-        value={this.props.value}
-        disabled={this.props.disabled}
-        readOnly={this.props.readOnly}
-        onChange={this.handleChange}
-        onFocus={this.props.onFocus}
-        onBlur={this.props.onBlur}/>
-    );
+    return <textarea className={className} {...props} />;
```
hturan: This is missing `onChange`

packages/cf-component-textarea/src/Textarea.js
```diff
@@ -16,22 +16,14 @@ class Textarea extends React.Component {
   }
 
   render() {
-    let className = 'cf-textarea';
+    let {className, ...props} = this.props;
+    
+    className = 'cf-textarea';
```
wyuenho: Now you've just overwritten the user supplied className :)

packages/cf-component-textarea/src/Textarea.js
wyuenho: No need. We can remove `handleChange` altogether as well.

packages/cf-component-textarea/src/Textarea.js
koddsson: Ah, I misread how the `cf-component-input` does this D:

wyuenho: _className :)

packages/cf-component-textarea/src/Textarea.js
koddsson: Won't that break for clients using this and expecting the value in the callback instead of the whole event?

packages/cf-component-textarea/src/Textarea.js
koddsson: I did it slightly differently but I can change it to match input if that's better.

packages/cf-component-textarea/src/Textarea.js
hturan: Yeah, this would be a semver major change.

koddsson: I'll add the onChange back in. I didn't see that it was different when I removed it.

wyuenho:
Should probably remove handleChange too, there's no use for it anymore.

PR #66, opened 2017-01-23
Add Travis CI support

### LINKED ISSUE #29: Set up CI
@jwineman said he wanted to get CI up and running. Travis should help with linting, but if we could set up saucelabs to run tests in many different browsers, that'll be really nice.

koddsson:
Fixes #29


koddsson:
This would need a admin to turn Travis on in the settings to work.

package.json
```diff
@@ -9,7 +9,7 @@
     "build": "gulp build",
     "dev": "gulp dev",
     "lint": "gulp lint",
-    "test": "gulp test",
+    "test": "gulp test --browsers PhantomJS",
```
koddsson: There didn't seem to be any way to overwrite the test command in Travis but I didn't think this is such a bad idea anyway.

koddsson:
@wyuenho Sounds good. Should I do it as part of this PR or should we get this in to have something and then iterate on it?

wyuenho:
You can get a SauceLab account try it on your own fork first right? PhantomJS has a lot of problems with DOM events.

koddsson:
Sure! I haven't had any problems with PhantomJS lately but I'll try the sauce labs thing as part of this PR then.

koddsson:
I'm just making sure that sauce labs works by testing this locally and I'll maybe ping for a re-review and merge :)

koddsson:
Ok so the problem I'm having on my local fork with my free tier of travis/sauce labs is that tests will timeout since I don't have enough concurrent VMs. Is this a problem I am gonna have to solve before merging or do we have 1 VM for each test we are gonna run in saucelabs?

cc/ @wyuenho

koddsson:
I'll try setting the concurrency setting to 8 and we can then just change it if we'd like later.

wyuenho:
I think you can launch just 1 browser instead of all 4 at the same time if you are just trying out.

koddsson:
This is taking waaaay more time than I can invest in this. If I revert this back into using phantomjs can we merge this and then add sauce labs in later. Then at least we'd have something™ instead of nothing™

cc/ @wyuenho

koddsson:
[Cool](https://travis-ci.org/cloudflare/cf-ui/builds/194897898)

PR #69, opened 2017-02-03
Adds an 'extra-wide' option to modal.

jculvey:
Wide just wasn't cutting it.

### DIFF
--- a/packages/cf-component-modal/example/basic/component.js
+++ b/packages/cf-component-modal/example/basic/component.js
@@ -23,7 +23,7 @@ class Component extends React.Component {
     super(props);
     this.state = {
       isModalOpen: false,
-      isWide: false
+      width: null
     };
     this.handleRequestOpen = this.handleRequestOpen.bind(this);
     this.handleRequestClose = this.handleRequestClose.bind(this);
@@ -39,7 +39,7 @@ class Component extends React.Component {
 
   handleWidthToggle() {
     this.setState((oldState) => ({
-      isWide: !oldState.isWide
+      width: oldState.width === 'wide' ? null : 'wide'
     }));
   }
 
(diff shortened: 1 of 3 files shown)


wyuenho:
How wide is extra-wide?

hturan:
If we're going this route can we do:

```js
if (this.props.width) {
  modalClassName += ` cf-modal--${this.props.width}`;
}
```

...to avoid when we need `super-wide` in the future?

jwineman:
Is this a pain to style at the page level?

jculvey:
@wyuenho it's 800px.

wyuenho:
Can you update the examples, readme and tests?

jculvey:
@jwineman yeah, unfortunately its plain impossible at the page level. Since the cf-ui Modal uses `react-gateway`, it gets re-parented to another DOM node at the bottom of the page, which makes it impossible to wrap it in something with a class we can target :/

jculvey:
I added a test, updated the example, and changed the functionality the way @hturan described.

I couldn't get the examples to work for me locally. It complains about missing semicolons in both cf-component-flex and cf-component-progress.

```bash
[13:19:38] Starting 'examplesDevVendor'...
[13:19:39] 'examplesBuildHtml' errored after 548 ms
[13:19:39] SyntaxError: Unexpected token, expected ; (1:6)
    at Parser.pp$5.raise (/Users/jculveyhouse/cf-repos/cf-ui/node_modules/babylon/lib/index.js:4246:13)
```

jwineman:
```
$ git pull origin master
$ npm run build
$ npm run bootstrap
$ npm run examples:build:internal
```
Works for me

jculvey:
K I ran the examples and they look fine. Anything else needed for this?

hturan:
> K I ran the examples and they look fine. Anything else needed for this?

Nope, let's get this in. I'm tired of my thin modals.

PR #70, opened 2017-02-06
Fixed bug where Checkbox onChange was missing context.

jwineman:
In 2300d36be98904ee8487863763af2e10a90c43e8 we had:
```
  handleChange = e => {
    this.props.onChange(e.target.checked);
  };
```
now we have:
```
  handleChange(e) {
    this.props.onChange(e.target.checked);
  }
```

So I added `bind(this)` so we continue passing the correct context and don't get error.


wyuenho:
Can you add a test for this?

packages/cf-component-checkbox/src/Checkbox.js
```diff
@@ -21,7 +21,7 @@ class Checkbox extends React.Component {
           name={this.props.name}
           value={this.props.value}
           checked={this.props.checked}
-          onChange={this.handleChange}/>
+          onChange={this.handleChange.bind(this)}/>
```
prestonp: The react docs recommend doing binds once in the constructor instead of every render. 

Although I don't actually know of any cases where performance is actually affected.. it probably doesn't hurt to make that optimization

wyuenho: Might as well get rid of `handleChange()` entirely and pass the `onChange` props directly to the input element without any binding. We'll up the major version number when we release.

jwineman:
Test added in 2e4220d95b06fb5b6ee5189bb23e3537cbdd1989

packages/cf-component-checkbox/test/Checkbox.js
```diff
@@ -1,8 +1,44 @@
...
 const assertEqualJSX = require('assert-equal-jsx');
 const Checkbox = require('../src/Checkbox');
+const {expect} = require('chai');
+const {render, unmountComponentAtNode} = require('react-dom');
+const TestUtils = require('react-addons-test-utils');
 
 describe('Checkbox', function() {
+  beforeEach(function() {
+    this.root = global.document.createElement('div');
+    global.document.body.appendChild(this.root);
```
wyuenho: This isn't run as a module so you don't need the `global` keyword

jwineman: 1996665ee9cc81a7f465d55c6eb7acc8d89d6f8f

packages/cf-component-checkbox/test/Checkbox.js
```diff
@@ -1,8 +1,44 @@
 const React = require('react');
 const assertEqualJSX = require('assert-equal-jsx');
 const Checkbox = require('../src/Checkbox');
+const {expect} = require('chai');
+const {render, unmountComponentAtNode} = require('react-dom');
+const TestUtils = require('react-addons-test-utils');
 
 describe('Checkbox', function() {
+  beforeEach(function() {
+    this.root = global.document.createElement('div');
```
wyuenho: There's still a global here.

packages/cf-component-checkbox/test/Checkbox.js
```diff
@@ -1,8 +1,44 @@
...
+    let called = false;
+    let onChange = () => called = true;
+
+    render(
+      <Checkbox label="false" name="name" value="value" checked="false" onChange={onChange}/>,
+      this.root
+    );
+
+    setTimeout(() => {
+      TestUtils.Simulate.mouseEnter(
```
wyuenho: Why can't you just simulate a change event directly?

jwineman: Because I'm a lazy copy/paster

jwineman: 0e2974e489f3f306115770101ffa2f016a9fd60d

packages/cf-component-checkbox/test/Checkbox.js
jwineman: wtf

jwineman: /me is an idiot

jwineman: e53092f5ac9a1854a22ac70e00e566cba42752b5

packages/cf-component-checkbox/test/Checkbox.js
```diff
@@ -1,8 +1,44 @@
...
 const assertEqualJSX = require('assert-equal-jsx');
 const Checkbox = require('../src/Checkbox');
+const {expect} = require('chai');
+const {render, unmountComponentAtNode} = require('react-dom');
+const TestUtils = require('react-addons-test-utils');
+
+var then = function (callback, timeout) {
+  setTimeout(callback, timeout > 0 ? timeout : 0);
+  return {then: then};
+};
```
jwineman: So it turns out simulating events with `React TestUtils.simulate` [isn't synchronous](http://brandonokert.com/2015/08/04/TestingInReact/#You_Cannot_Check_State/Render_or_Simulate_Right_After_Simulating_an_Event).  The guy who wrote that article also provides some [utility patterns](http://brandonokert.com/2015/08/04/TestingInReact/#Utility_Patterns) which might be useful for us to extract into a lib so that we're consistent with how we test this stuff. 

In `<Notifications/>` I've seen other home grown methods for dealing with async testing scenarios.  I'm sure they exist else where.

wyuenho: I think we are using enzyme in a couple of places already. Enzyme's simulate will grab the onClick handler and call it directly. This may make the tests more straightforward conceptually.

http://airbnb.io/enzyme/docs/api/ShallowWrapper/simulate.html

jwineman: Enzyme is a million times better.

packages/cf-component-checkbox/test/Checkbox.js
```diff
@@ -1,8 +1,21 @@
...
 const Checkbox = require('../src/Checkbox');
+const { expect } = require('chai');
+const { shallow } = require('enzyme');
 
 describe('Checkbox', function() {
+  it('should handle onChange', function() {
+    let called = false;
+
+    let wrapper = shallow(
+      <Checkbox onChange={ () => called = true } />
```
wyuenho: no space after { and before }

packages/cf-component-checkbox/src/CheckboxGroup.js
```diff
@@ -34,7 +34,7 @@ class CheckboxGroup extends React.Component {
             name={option.name}
             value={option.value}
             checked={includes(this.props.values, option.value)}
-            onChange={val => this.handleChange(option.value, val)}/>;
+            onChange={e => this.handleChange(option.value, e.target.checked)}/>;
```
tajo: @jwineman You asked me because of this? If you are doing micro-optimizations, using immutable data structures and pure components that implement `shouldComponentUpdate()`, this is not a good solution because it [creates a new function with a new identity](https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f#.ml95bv4bt), so the `render()` of Checkbox is called every single time. But we don't do that, so it doesn't really matter and you need to pass that option param somehow.

Anyway, is there some reason why we don't use functional components?

wyuenho: >  is there some reason why we don't use functional components?

We'd like to preserve the ability to use refs if we need to.

I agree this onChange handler in CheckboxGroup is weird, but that's a separate PR. cf-ui has a number of these  arrow functions inlined into JSX that delegate to some class method handlers with weird signatures.

tajo: > We'd like to preserve the ability to use refs if we need to.

I mean we would use always both. You still need class components in case of lifecycle hooks (`componentDidUpdate()`) but if you can (vast majority of cases), it's nicer to use functional components.

wyuenho: Meh, it's not going to buy us much until React Fiber lands. Before that, it's just a nicer syntax.

PR #72, opened 2017-02-07
Gimme all the flex props

wyuenho:
So we can use this component to do complex layout semantically rather than sprinkling divs with non-prefixed inline styles all over the place.

### DIFF
--- a/packages/cf-component-flex/src/Flex.js
+++ b/packages/cf-component-flex/src/Flex.js
@@ -1,16 +1,30 @@
 const React = require('react');
 const {PropTypes} = React;
+const prefixAll = require('inline-style-prefixer/static');
+
+
+function styles (props) {
+  return {
+    flexDirection: props.direction,
+    flexWrap: props.wrap,
+    justifyContent: props.justifyContent,
+    alignItems: props.alignItems,
+    alignContent: props.alignContent
+  };
+}
 
 class Flex extends React.Component {
   render() {
-    let className = 'cf-flex';
+    let className = 'cf-flex', style;
 
     if (this.props.spacing) {
       className += ' cf-flex--' + this.props.spacing;
+    } else {
+      style = prefixAll(styles(this.props));
     }
 
     return (
-      <div className={className}>
+      <div className={className} style={style}>
         {this.props.children}
       </div>
     );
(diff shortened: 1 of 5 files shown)


packages/cf-component-flex/src/Flex.js
```diff
@@ -1,16 +1,41 @@
...
+        const [cssProp, prop] = pair;
+        if (prop) {
+          style[cssProp] = prop;
+        }
+      });
     }
 
     return (
-      <div className={className}>
+      <div className={className} style={prefixAll(style)}>
```
prestonp: Why not just pass through props.style? It seems like a lot of extra mental overhead to define this new stylesheet interface

wyuenho: I don't want to open up styles unrelated to flexbox to be passed to a flex component

wyuenho: @prestonp can you explain what you mean by mental overhead?

prestonp: Yeah, there's the interface to style that react has, and this introduces another layer of css through _component properties_. I think avoiding another layer would make things easier

prestonp: What about filtering the flexbox stuff on props.style, something like

```
_.pick(props.style, 'spacing', 'direction', 'wrap', ...)
```

But the es6 equivalent of _.pick

wyuenho: I always thought `style` is a `react-dom` prop instead of a component prop convention. By listing the accepted flex props out, I can use PropTypes to check them, and they are shorter, otherwise you'll have to annoyingly prepend `flex` to many CSS properties because DOM.

prestonp: I'd rather lean on the browser to validate the css instead of using PropTypes. In general though I think this PR looks good.

wyuenho: When the values are rendered, devtool only tell you the value is valid instead of why it's invalid. I mean, this isn't a new pattern, it's just instead of appending a className, we are inserting inline styles because flexbox.

tajo: This is a perfect example where CSS in JS will shine. The component API is good / makes sense but the implementation itself can be much nicer with Fela (not using inline styles, automatic prefixing...).

wyuenho: Speaking of fela, I've significantly cleaned up the PR using similar styles, can you look again?

packages/cf-component-flex/src/Flex.js
```diff
@@ -1,16 +1,30 @@
...
 
 class Flex extends React.Component {
   render() {
-    let className = 'cf-flex';
+    let className = 'cf-flex', style;
 
     if (this.props.spacing) {
       className += ' cf-flex--' + this.props.spacing;
+    } else {
+      style = prefixAll(styles(this.props));
```
tajo: Much nicer. 👍

PR #78, opened 2017-02-24
Prettier

tajo:
- The whole codebase is fixed by Prettier
- `npm run lint` is using rules from Prettier
- `npm run prettier` fixes everything
- .sublime-project added, so prettier is triggered on auto-save


prettier.js
```diff
@@ -0,0 +1,59 @@
...
+    'build/**',
+    'coverage/**',
+    'packages/*/lib/**',
+    'docs/bundle.js',
+    'docs/vendor.js'
+  ]
+});
+
+files.forEach(file => {
+  fs.readFile(file, 'utf8', (err, data) => {
```
jwineman: These rules are from Tony, please don't bike shed them.

package.json
```diff
@@ -11,6 +11,7 @@
     "lint": "gulp lint",
     "test": "gulp test --browsers PhantomJS",
     "updated": "lerna updated --only-explicit-updates",
+    "prettier": "node prettier.js",
```
jwineman: I'd rather call this `format` to be more consistent with our `package.json` commands else where but as long as it exists I'm good.

PR #79, opened 2017-02-24
Fixed duplicate active props. Now uses activeTab

manatarms:
Tabs had duplicate active tabs props of which one was always undefined, which meant it always entered the if loop.


jwineman:
Sachin merged this on accident because he thought he was on his fork.  Please still review it.

I just put in a request to require 1 approval on CF-UI to keep stuff like this from happening in the future. Your PR looks okay, I'll hold of doing npm publish until people have a chance to look at it to make sure we aren't breaking anyone's use case.

manatarms:
Sorry about this! I thought I was on my own fork when I merged it.

prestonp:
Overall looks good. I only saw one reference to it internally and it was using `activeTab` not `active`.

One concern that could be fixed later is cf-component-card exports CardDrawer which uses `active` 

https://github.com/cloudflare/cf-ui/blob/master/packages/cf-component-card/src/CardDrawers.js#L74

This is really nitpicky, but I think we should just use `active` everywhere

koddsson:
Did someone release this?

jwineman:
Not yet.  @manatarms make another PR to switch to `active` to be consistent like @prestonp said then we'll release.

manatarms:
@jwineman Switching to active might break all current instances of the tab component, because most of the ones I've seen use activeTab instead of active, including the example. Should I comb the code base for uses before changing this?

jwineman:
We're release it as a major release to indicate its a breaking change.  We'll have to go through our code and update every instance of the component but it won't be difficult to find and replace.

PR #80, opened 2017-02-24
Remove npm owners + scripts

tajo:
This should not be needed since we have now a npm organization and teams.


jwineman:
Assuming lerna doesn't depend on any of this it looks good.

PR #81, opened 2017-02-24
Support centered toggles

jroyal:
image

### DIFF
--- a/example/styles.css
+++ b/example/styles.css
@@ -2062,6 +2062,10 @@ a.cf-link--disabled {
   color: #888;
 }
 
+.cf-toggle--centered {
+  margin: 0 auto;
+}
+
 /**
  * Tooltip
  */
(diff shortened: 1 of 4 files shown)


jwineman:
I would vote against this because I prefer the positioning of stuff to happen in the page css.  You should be able to center the `<Toggle/>` with:

```
<Text align="center"><Toggle .../></Text>
```

Eventually when we switch to Fela you should be able to customize the theme to `align: center;`

What is your `<Toggle/>` inside of?  On partners we have an [example](https://github.com/cloudflare/cloudflare-plugin-frontend/blob/master/src/containers/DNSRecordEditor/DNSRecordEditor.js#L49) of `<Toggle/>` being centered inside of a `<TableCell\>` without any extra CSS.

screen shot 2017-02-24 at 2 28 23 pm

jroyal:
I am going to close this, because I was able to do it with an existing css sheet.

Issue #83, opened 2017-02-27
Change activeTab to active in cf tabs component


manatarms:
Creating this issue to track the PR for moving to active from activeTab prop in the cf-component-tabs package.

So far I've only found the following use of activeTab in www-next.
1) containers/analytics/billboard/RateLimitingCard.js:

wyuenho:
This is fixed right?

jwineman:
Yea in

manatarms:
Yup. It's fixed but to keep the props consistent we're renaming the activeTab prop to active. This will break stuff so releasing it as a major change.

jwineman:
Closing since the PR is merged. @manatarms knows how to find the affected components when he bumps the version after it gets released.

PR #84, opened 2017-02-27
Breaking change active tab

manatarms:
Moving from activeTab to tab in this PR. I've also created an issue #83  to track anything we need to change.


wyuenho:
Really?

manatarms:
@koddsson Please publish this as well when you publish #85. Thanks :)

PR #85, opened 2017-02-27
Send all props down to react-select.

koddsson:
I could also c/p the props from the react-select documentation if we'd prefer that :D

### DIFF
--- a/packages/cf-component-select/src/Select.js
+++ b/packages/cf-component-select/src/Select.js
@@ -4,55 +4,20 @@ const ReactSelect = require('react-select');
 
 class Select extends React.Component {
   render() {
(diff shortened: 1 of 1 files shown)


packages/cf-component-select/src/Select.js
```diff
@@ -4,56 +4,21 @@ const ReactSelect = require('react-select');
...
-          placeholder={this.props.placeholder}
-          disabled={this.props.disabled}
-          cache={this.props.cache}
         />
       </div>
     );
   }
 }
 
-Select.propTypes = {
```
koddsson: [react-select already does this](https://github.com/JedWatson/react-select/blob/master/src/Select.js#L47)

PR #90, opened 2017-03-02
Don't render tab panels unless they are clicked

koddsson:
`cf-component-tabs` would render all the tabpanel components when a tab component would get rendered. This has the potential of loading a bunch of data over the network that the user will never see.

By changing it this way we are able to only request the data needed to render the tabs the user actually clicks on

### DIFF
--- a/packages/cf-component-tabs/src/Tabs.js
+++ b/packages/cf-component-tabs/src/Tabs.js
@@ -8,7 +8,6 @@ const find = (list, condition) => {
   list.forEach(element => {
     if (condition(element)) {
       foundElement = element;
-      return;
     }
   });
   return foundElement;
--- a/packages/cf-component-tabs/src/Tabs.js
+++ b/packages/cf-component-tabs/src/Tabs.js
@@ -92,7 +92,7 @@ Tabs.propTypes = {
       label: PropTypes.string.isRequired
     })
   ).isRequired,
-  children: PropTypes.node
+  children: PropTypes.arrayOf(PropTypes.element).isRequired
 };
 
 Tabs.childContextTypes = {
--- a/packages/cf-component-tabs/src/Tabs.js
+++ b/packages/cf-component-tabs/src/Tabs.js
@@ -64,7 +64,9 @@ class Tabs extends React.Component {
             })}
           </ul>
         </Viewport>
-        {this.props.children}
+        {this.props.children.find(child => {
+          return child.props.id === this.props.active;
+        })}
       </section>
     );
   }
--- a/packages/cf-component-tabs/src/Tabs.js
+++ b/packages/cf-component-tabs/src/Tabs.js
@@ -3,6 +3,17 @@ const { PropTypes } = React;
 const Viewport = require('cf-component-viewport');
 const Select = require('cf-component-select');
 
+const find = (list, condition) => {
+  let foundElement = undefined;
+  list.forEach(element => {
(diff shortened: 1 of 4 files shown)


packages/cf-component-tabs/src/TabsPanel.js
```diff
@@ -5,13 +5,14 @@ class TabsPanel extends React.Component {
   render() {
     const selected = this.context.active === this.props.id;
 
+    console.log(`Rendering tab ${this.props.id}`);
```
koddsson: Whoops!

jwineman:
Can you write a test ensuring the other ones aren't rendered?

packages/cf-component-tabs/src/Tabs.js
```diff
@@ -64,7 +64,9 @@ class Tabs extends React.Component {
             })}
           </ul>
         </Viewport>
-        {this.props.children}
+        {this.props.children.find(child => {
```
hturan: Be careful when dealing with `this.props.children`, it's perfectly valid for it to be a string. See https://facebook.github.io/react/docs/react-api.html#react.children for more info.

hturan: `Array.find` is ES6. Are we polyfilling this?

koddsson:
@jwineman absolutely!

packages/cf-component-tabs/src/Tabs.js
koddsson: 1. Could I just change the proptype to be a array of elements?
2. Not sure. I'll check.

koddsson: 2. So we are using babel and `Array.prototype.find(...)` is in `core-js`?

koddsson:
I have a hard time beliving the tests are actually running on staging since I just ran the tests and this PR breaks 'em.

packages/cf-component-tabs/test/Tabs.js
```diff
@@ -39,18 +39,8 @@ describe('Tabs', function() {
...
           role="tabpanel"
-          aria-labelledby="1"
-          aria-hidden="true"
-          style={{ display: 'none' }}
-        >
-          One
-        </div>
-        <div
-          className="cf-tabs__panel"
-          role="tabpanel"
```
koddsson: @jwineman I didn't add specific tests since I feel like this is covered by the "will render" test. I can try to make a more explicit one if you'd prefer.

packages/cf-component-tabs/src/Tabs.js
```diff
@@ -3,6 +3,17 @@ const { PropTypes } = React;
 const Viewport = require('cf-component-viewport');
 const Select = require('cf-component-select');
 
+const find = (list, condition) => {
+  let foundElement = undefined;
+  list.forEach(element => {
+    if (condition(element)) {
+      foundElement = element;
+      return;
```
hturan: I don't think you can short-circuit `forEach` like this, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Description.

koddsson: Good catch! I just removed the return since the list realistically isn't gonna be much bigger than 10 elements.

koddsson:
@jwineman Can I merge this?

jwineman:
image

Issue #82, opened 2017-02-27
Missing react-fela in `cf-component-button` 4.0.0


koddsson:
This code:

```js
import React from 'react';
import {render} from 'react-dom';
import {Button} from 'cf-component-button';

const App = () => {
  return <Button>Click me</Button>
}

render(<App />, document.querySelector('body'));
```

results in this error
`stderr: Error: Cannot find module 'react-fela' from '/tmp/cf-component-button117127-29294-18e4ou5/node_modules/cf-component-button/node_modules/cf-style-container/lib'`

koddsson:
@tajo Do you know how we can fix this?

hturan:
It looks like all of our `cf-component-button` tests were commented out as part of #77, meaning this wasn't caught in CI.

tajo:
Solved with https://github.com/cloudflare/cf-ui/pull/91

PR #91, opened 2017-03-02
CSS in JS Polishing

tajo:
- new package `cf-const-style`, it contains global variables (global theme)
- local themes (that come with components) are now functions that get the global theme through parameter so they can reference it (and not just overwrite it)
- `cf-style-container` has a new HOC called `applyTheme`, it's a handy alternative to ThemeProvider since it creates a styled component that can be repeatedly used without need of ThemeProvider wrapping everything
- most of these changes are result of my work on the internal cf-ui (style wrappers for this project), so everything can play nicely together ... so check stash/www/cf-ui as well


jwineman:
Is it worth writing tests or is this still going to change significantly?

tajo:
@jwineman It is! Afaik, the HTML markup should be exactly the same as for old cf-component-button. The only new thing is different className. For the purpose of deving, we use [monolithic enhancer](https://www.npmjs.com/package/fela-monolithic) so fela assigns only one className to each component. The className is a hash of all style properties. I think this would be great for snapshot tests as well! You'll break them if you change the HTML markup or if you change styles. **So not only our styles are now using semver but they will be covered by tests too!**

jwineman:
Are you saying our existing tests pass for these changes?

tajo:
They would fail on `class="cf-button"` ...

jwineman:
So lets update the tests to pass as part of this PR.

tajo:
> So lets update the tests to pass as part of this PR.

So, I'll fix tests in a different PR where I'll introduce Jest because we really need automized snapshot tests here.

packages/cf-style-const/src/variables.js
```diff
@@ -0,0 +1,41 @@
+import cssColorFunction from 'css-color-function';
+
+const color = input => cssColorFunction.convert(`color(${input})`);
+
+const t = {};
```
hturan: Super minor thing, but is there any reason why we're not writing this as:

```js
export default {
  fontSize: '15px',
  ...etc
};
```

hturan:
Looking good! I'm guessing this fixes #82 too?

packages/cf-style-const/src/variables.js
tajo: Because that way you can't self reference these values which we do a lot. We could be more strict and define constants at the top of file. Or we could even use currying. Not sure what annoys me less tho. :)

hturan: Cool, makes sense! I missed the `fontWeight` assignment.

packages/cf-component-button/package.json
```diff
@@ -10,7 +10,8 @@
   },
   "dependencies": {
     "cf-style-container": "^0.0.2",
-    "react": "^0.14.2 || ^15.0.0-0"
+    "css-color-function": "^1.3.0",
+    "react": "^15.0.0-0"
```
sejoker: is it intentional to drop support of 0.14? I think support of last 2 major versions is a good strategy.

jwineman: Unless fela doesn't support 14 I agree we should maintain backwards compatibility.

packages/cf-style-const/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "cf-style-const",
+  "description": "Cloudflare Style Constants",
+  "version": "0.0.1",
```
sejoker: why not 1.0.0? I think semver rules before 1.0 are quite bad.

packages/cf-component-button/src/ButtonTheme.js
```diff
@@ -1,26 +1,85 @@
...
+    `${baseTheme.dangerBackground} lightness(${baseTheme.colorOffsetDark})`
+  );
+
+  // hover colors
+  t.defaultHoverColor = t.defaultColor;
+  t.primaryHoverColor = t.primaryColor;
+  t.successHoverColor = t.successColor;
+  t.warningHoverColor = t.warningColor;
+  t.dangerHoverColor = t.dangerColor;
+  return t;
```
sejoker: do we want Object.freeze(t)? Will prevent changing values inside the object.

Issue #92, opened 2017-03-08
cf-component-tooltip can't be server-side rendered


tajo:
It uses "tether-tooltip" and that library touches DOM all over the place. The new styleguide is going to be server-side rendered so this is breaking my styleguide build currently. I'm going to remove it from the styleguide and we should replace "tether-tooltip" with something better. 

[react-tooltip](https://github.com/wwayne/react-tooltip) looks pretty good and we could preserve similar (same?) API.

I tried many hacks and tricks how to bypass it on server but it's not worth of effort just because of one single component.

**In general, all our cf-ui component should be compatible with server side rendering.**

sejoker:
I used react-tooltip in the past without any issues. The app supported server-side rendered.

koddsson:
Is this a think we want to work on?

I don't think server-side rendering is actively being worked in the team for now so I'd like to focus more on stuff we are actually going to do and bring this back up later if we ever wanna do server side rendering.

Ok if I close this with maybe a label of "Possible future work" so it's easy to find again later?

tajo:
> Is this a think we want to work on?

**Yes.** 

We already use server-side rendered cf-ui components for [api docs](https://api.cloudflare.com) and [styleguide](https://cloudflare.github.io/cf-ui/). So this is a blocker preventing us to use cf-component-tooltip in both of these projects. **There is no excuse to throw server-side compatibility out of the window**.

koddsson:
I mean the argument could be made that we're spending time on issues like this where we are making sure that we are able to server-side rendering even though we don't use it in the dashboard. I'd argue that we aren't throwing it out the window but rather just not taking up time implementing it when we aren't really using it.

Why are we server-side rendering in the styleguide and api docs?

tajo:
> taking up time implementing it 

It doesn't really take an extra time. It's just not doing bad things in our components (using global stuff as `window` etc).

> Why are we server-side rendering in the styleguide and api docs?

- **faster loader time**, the page doesn't wait for JS bundle, parsing, executing...
- better for search engines
- we could ship docs without JavaScript and it would still work pretty good

koddsson:
Is this being worked on or is this just a feature request for the future?

toekneestuck:
Moved internal.

Issue #73, opened 2017-02-17
Jest for testing


tajo:
The current tests are... slow. Also, [snapshot tests](https://facebook.github.io/jest/docs/snapshot-testing.html) would be great for our use-case (we could cut the size of tests in half). 

So, how do you feel about leaving Karma/assertEqualJSX and using Jest + node.js? Running tests in real browsers is nice but [the speed is everything](https://github.com/facebook/jest/issues/139#issuecomment-229277654).

jculvey:
👍

handrews:
@tajo is there data on how robust these snapshot tests are?  Maintaining screenshot-based tests is frequently a nightmare, but I am not familiar with this tool and it's been a while since I dealt with the problem.  Does it require pixel-perfect comparisons or is there a fuzz factor?

Generally, there is a need for some browser-specific testing but certainly not everything needs to go through a real browser.

tajo:
@henry It's not doing real screenshots and comparing pixels. 

It takes a React Component, renders the HTML markup, saves it and in the next run it compares if the markup is unchanged. So it basically works the same way as our current tests except there is no need to manually write the markup that should be generated.

handrews:
@tajo oh awesome.  I spent years designing QA automation systems and things that seem like screenshot-based tests give me anxiety attacks ;-)  There always seems to be someone who thinks they can make it work... (I've never seen it truly scale successfully)

jwineman:
> There always seems to be someone who thinks they can make it work
Inadvertently describes all software development :)

wyuenho:
I'd love to know how much time we spend in babel and webpack before jumping to the conclusion that karma is slow. Opening a window, sending code to it, opening sockets, running tests inside the browser and sending the results back from the local network, none of these sound very slow to me. Profiling data please.

wyuenho:
To give you some perspective. For backgrid, running 195 tests on 3 browsers (555 total) simultaneously took about 8 seconds.

tajo:
@wyuenho The whole `npm run test` takes 1min 30s (on the latest quad core i7). That's bad.

You are right. The problem is not Karma itself. That part takes about last 10s (although I am sure it could be even better with Jest since Karma doesn't parallelize on the component level).

[...]

Issue #67, opened 2017-01-27
Upgrade superagent


wyuenho:
We are using superagent 1.x in cf-util-http, superagent is at 3.x now. Should probably upgrade before it's too late.

jwineman:
This is kind of frustrating because Jest swallows `console.log()` if the test errors out.  Since sinon's fake server is failing _inside_ the test itself its hard to tell whats broken.

tajo:
Superagent replaced by fetch.

Issue #23, opened 2016-11-28
Migrate to Webpack


wyuenho:
So we can remove the bulk of hand written gulp stuff, get a faster build process and application code can take advantage of Webpack's treeshaking.

https://webpack.js.org/guides/author-libraries/

tajo:
Done with https://github.com/cloudflare/cf-ui/pull/93

PR #93, opened 2017-03-14
Modernize cf-ui

tajo:
## Changes
- Gulp is gone. Webpack rules everything now. https://github.com/cloudflare/cf-ui/issues/23
- The style-guide is now an app that's using webpack so you can use `npm start` and enjoy hot module replacement. You can build `/docs` through `npm run build:styleguide`.
- The style-guide is server-side rendered (including Fela styles). There was a problem with [cf-component-tooltip](https://github.com/cloudflare/cf-ui/issues/92) so it's not visible in the style guide for now.
- CommonJS modules replaced by ES6 modules.
- New build targets: ES6, UMD and UMD minified. Webpack2 can now use ES6 modules to do tree-shaking. CommonJS build is now done via Webpack as well.
- Karma, Mocha, Chai, Browserify... all gone. All tests are using Jest. In rare cases where the DOM is needed, we use JSDom now. https://github.com/cloudflare/cf-ui/issues/73
  - I had some weird problems with cf-util-http and cf-util-http-poll tests. They are skipped now. @jwineman promised to fix them :-). It's related to https://github.com/cloudflare/cf-ui/issues/67.
- All READMEs udpated.

**I didn't change functionality of any component. Everything's related only to the build process, style guide and testing.**

There are probably some small errors in the style guide. Let's fix them later!


terinjokes:
:cold_sweat:

discussions/no-customizing-classnames.md
```diff
@@ -1,58 +0,0 @@
-# No Customizing `className`'s
```
jwineman: Why are we deleting this?

discussions/no-es6-modules.md
```diff
@@ -1,53 +0,0 @@
-# No ES6 Modules
```
jwineman: I might update this rather than deleting it, explaining why we chose webpack2 etc.

CONTRIBUTING.md
```diff
@@ -18,47 +18,46 @@ $ npm run bootstrap
...
-$ git checkout master
-$ git pull origin master --rebase
-$ ./node_modules/.bin/check-installed-dependencies
-$ npm run build
-$ cd packages/[[package-name]]
-# edit version inside package.json
-$ git tag -a "[[version]]"
-$ npm publish
-$ git push origin master --follow-tags
+npm run publish
```
sejoker: if someone wants to publish single/multiple packages is the 'npm run publish' will be used in both cases?

discussions/functional-testing-containers.md
```diff
@@ -21,7 +21,7 @@ functions that take state and render UI.
 A unit test for state => ui looks something like this:
 
 ```js
-it('should render the list of items', function() {
+test('should render the list of items', function() {
```
sejoker: could update using `() => {}` syntax

package.json
```diff
@@ -1,120 +1,88 @@
...
-    "assert-equal-jsx": "^1.0.3",
-    "autoprefixer": "^6.3.6",
+    "autoprefixer": "^6.7.6",
+    "babel-cli": "^6.23.0",
     "babel-core": "^6.18.0",
     "babel-eslint": "^7.1.1",
-    "babel-generator": "^6.18.0",
-    "babel-loader": "^6.3.2",
+    "babel-jest": "^19.0.0",
+    "babel-loader": "7.0.0-alpha.3",
```
sejoker: are you sure it is a good idea using alpha version?

packages/cf-builder-pagination/example/basic/component.js
```diff
@@ -1,6 +1,5 @@
-const React = require('react');
-const { render } = require('react-dom');
-const { PaginationBuilder } = require('../../src/index');
+import React from 'react';
```
sejoker: we change from common.js to es6 modules syntax. Is it not enough to bump package version of all components?

discussions/no-customizing-classnames.md
sejoker: I guess because we use Fela, it will generate className automatically

tajo: Well, it's still true but we need to expose classNames so Fela can inject styles into the components. Eventually, I'm going to write some new pieces explaining the whole Fela thing.

discussions/no-es6-modules.md
tajo: Yea. I'll write something but later.

discussions/no-customizing-classnames.md
jwineman: I think the whole no `className` thing still applies.  Customizing the classes through Fela is not the same as customizing it directly.  Its probably worth explaining.

.babelrc
```diff
@@ -1,4 +1,74 @@
 {
-  "presets": ["es2015", "react"],
-  "plugins": ["transform-object-rest-spread"]
+  "presets": [
+    ["es2015", { "modules": false }],
```
tajo: This PR changes everything to ES6 modules. Now, there's webpack2 that can handle them so there is no need for babel transformation which was main reason why not to use them before.

CONTRIBUTING.md
tajo: yes, learna has a nice interface to select what and how to publish

package.json
tajo: yea, because of this https://github.com/babel/babel-loader/pull/391

packages/cf-builder-pagination/example/basic/component.js
tajo: Not sure what you mean by this.

discussions/no-customizing-classnames.md
tajo: It definitely does, I just don't want to confuse people with "no classNames" article and then exposing classNames everywhere without explaining Fela bits.

discussions/functional-testing-containers.md
tajo: Fixed.

build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
tajo: Done.

discussions/no-customizing-classnames.md
jwineman: I agree the current article is confusing given our current implementation.  We don't have to do it in this PR but we should talk about separation of concerns at some point.  The idea here is that the component is responsible for functionality and Fela is responsible for style, and modifying `className` without Fela violates that.

packages/cf-builder-pagination/example/basic/component.js
hturan: Changing the module syntax for our imports doesn't represent a SEMVER incompatible API change, as far as I'm aware.

tajo: The actual `/lib` output should be identical. The main index.js is still commonjs module. But now we add `/es` target where ES6 modules are preserved and webpack2 or rollup can leverage that. Also `/dist` target for UMD build - I'm going to put them on cdnjs.com.

PR #94, opened 2017-03-14
Add a creatable parameter to enable Creatable select components to be…

marksteyn:
… created

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -67,7 +67,7 @@
   "env": {
     "commonjs": {
       "plugins": [
-        ["transform-es2015-modules-commonjs", { "loose": true }]
+        ["transform-es2015-modules-commonjs"]
       ]
     }
   }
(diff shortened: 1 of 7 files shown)


packages/cf-component-select/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "react": "^0.14.2 || ^15.0.0-0",
-    "react-select": "1.0.0-rc.2 || 1.x"
+    "react-select": "1.0.0-rc.3 || 1.x"
```
jwineman: I realize it was RC 2 before but shouldn't we be depending on a more stable version?  Does the functionality we need exist in 1.0.0+?

tajo:
Please rebase.

wyuenho:
2 things concerning here. I've tested this on the latest nvm, node 6.10.0, and npm 3 and 4.

1. `["transform-es2015-modules-commonjs", { "loose": true }]` should make transpiling the src files fail during test runs, somehow CI doesn't show this.
2. CI shows the same test result as if `loose: false`, but a number of us can't reproduce this on our local machines.

PR #98, opened 2017-03-16
cf-component-text to fela migration

sejoker:


packages/cf-component-text/src/TextTheme.js
```diff
@@ -0,0 +1,19 @@
...
+  'textTransform--capitalize': 'capitalize',
+  'textTransform--capitalize:first-word': 'capitalize',
+  'textTransform--lowercase': 'lowercase',
+  'textTransform--uppercase': 'uppercase',
+  'color--info': '#00a9eb',
+  'color--success': '#68970f',
+  'color--warning': '#fca520',
+  'color--error': '#ff3100',
+  'color--muted': '#gray'
+});
```
jwineman: We can move the colors here: https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-const/src/variables.js

sejoker: TextTheme styles are specific to our examples styles. Do we want them to be part of the cf-style-const?

tajo: Yes please. Also, I would rename them to `colorWarning`, `colorError` ... etc. So it's JS compatible.

sejoker: I can rename to `colorwarning`, `colorerror` easily, but camelCase format will require more work, is it worth the effort?

tajo: Yea, we should stay consistent and use camelCase everywhere. It's easy to do it now than later.

packages/cf-component-text/src/TextTheme.js
```diff
@@ -1,19 +1,24 @@
...
+    lineHeightSmall: 1.3,
+    textAlignStart: 'left',
+    textAlignCenter: 'center',
+    textAlignJustify: 'justify',
+    textAlignEnd: 'right',
+    textTransformCapitalize: 'capitalize',
+    'textTransformCapitalize:first-word': 'capitalize',
+    textTransformLowercase: 'lowercase',
+    textTransformUppercase: 'uppercase',
+    colorInfo: '#00a9eb',
```
jwineman: `const`?

sejoker: @jwineman my bad, updated PR

PR #101, opened 2017-03-17
Remove style guide and use relative paths for tests

tajo:
Another 100+ files PR but the vast majority of changes is just a replacement of absolute paths in the tests. I was using babel to remap all packages but it turns out that it is a disaster if you start "touching" this repo from other repos.

- the style guide has been removed, now it lives in our stash/www/cf-styleguide - it's a new shiny app that can automagically run internal and external styles, no symlinks, no extra steps... the only assumption is that all cf-* repos are in the same folder, it's even using postcss as a source of internal styles

That's it. Please LGTM asap.

**I'm going to document cf-styleguide since it uses a lot of webpack magic and write a private blog post as well.**


hturan:
-4k lines, that's what I like to see! Good stuff.

So is the plan here to make our cf-ui documentation private? I'm kind of confused around what `/docs` is, what `/styleguide` is, what `cf-ui-private` is and what `cf-styleguide` is. I'm guessing that you'll clear it all up in the upcoming blog post, but it does make it hard to review this without knowing the overall goal of what we're trying to do here.

package.json
```diff
@@ -15,17 +15,12 @@
...
     "build:umd": "cross-env BABEL_ENV=commonjs NODE_ENV=development webpack --config webpack.babel.js",
     "build:umd:min": "cross-env BABEL_ENV=commonjs NODE_ENV=production webpack --config webpack.babel.js",
-    "build:styleguide": "rimraf docs && cross-env BABEL_ENV=commonjs NODE_ENV=production webpack --config styleguide/config.prod.babel.js && rm docs/static-*",
-    "lint": "eslint packages/*/src/**/*.js packages/*/example/**/*.js packages/*/test/**/*.js styleguide *.js || true",
+    "lint": "eslint packages/*/src/**/*.js packages/*/example/**/*.js packages/*/test/**/*.js || true",
     "test": "cross-env BABEL_ENV=commonjs jest",
     "updated": "lerna updated --only-explicit-updates",
     "prettier": "node prettier.js",
-    "clean": "lerna clean --yes && rimraf .jest dist",
+    "clean": "lerna clean && rimraf dist",
```
jwineman: `rimraf .jest` needs to be part of `npm run clean` to fix the jest cache issues we were having earlier this week.

jwineman: also I don't know what `--yes` does but did you remove it on purpose?

packages/cf-builder-table/test/tableActions.js
```diff
@@ -1,8 +1,12 @@
...
+import { createMockStore } from '../../cf-test-store/src/index';
+import cf from '../../cf-pony/src/index';
+import {
+  tableActions,
+  TableActionTypes
+} from '../../cf-builder-table/src/index';
 
-test('.flashRow() should dispatch actions for flashing the row', done => {
+// this test is flaky, in 1/4 cases it just time outs
+test.skip('.flashRow() should dispatch actions for flashing the row', done => {
```
sejoker: I run `npm test -- --testPathPattern cf-builder-table` 10 times without flaky results. Let's hold on with skiping it

sejoker:
I am ok with changing pathes to relative, but as @hturan mentioned moving styleguide into private repo is something I want to discuss before making a change. It quite important to have open styleguide and within same repo, much easier to test.

package.json
tajo: Good catch. My mistake during the rebase.

tajo: Fixed.

packages/cf-builder-table/test/tableActions.js
tajo: Yea, it seems ok now. Fixed.

tajo:
@hturan @sejoker I'm going to document everything today. The reason why we are removing the styleguide from this repo is to streamline its development. We need to use it for our private styles and private components as well and we don't want to maintain two different versions. 

It's pretty hard to have a general styleguide app that could be open sourced right now (in sense of making no assumptions about our internal projects). The priority now is to provide the best DX and seamless integration with cf-ui-private and cf-www-next without running some crazy .sh scripts. That doesn't mean we can't bring it back to public later.

> It quite important to have open styleguide and within same repo, much easier to test.

I agree that the styleguide is critical for deving and testing of cf-ui/cf-ui-private components. It's one of its primary goals - to help developers maintain our React components. However, it doesn't have to be in the same repo thanks to some webpack magic.

^^ We were discussing this with @jculvey @jwineman and @toekneestuck.

packages/cf-builder-table/test/TableBuilder.js
```diff
@@ -101,6 +110,6 @@ test('should support flashes', () => {
     </Provider>
   );
   store.dispatch(tableActions.flashRow('test-table', '1', 'success'));
-  const row = wrapper.find(TableBody).find(TableRow).at(0);
-  expect(row.prop('type')).toBe('success');
+  const row = wrapper.find('tbody').find('tr').at(0);
```
wyuenho: Why?

packages/cf-builder-pagination/test/PaginationBuilder.js
```diff
@@ -86,7 +85,7 @@ test('should should call onPageChange when clicking another page', () => {
     />
   );
 
-  const items = wrapper.find(PaginationItem);
+  const items = wrapper.find('li');
```
wyuenho: ?

packages/cf-builder-table/test/TableBuilder.js
tajo: Enzyme was giving me this error:

```
Method “props” is only meant to be run on a single node. 0 found instead.
```

Not sure why so I'm testing against the markup.

packages/cf-builder-pagination/test/PaginationBuilder.js
tajo: Explained above. Not sure why Enzyme doesn't like this anymore but the is still legit (testing the same functionality).

PR #102, opened 2017-03-17
Updates Page and PageHeader components to use Fela

manatarms:
Few things to Note in this PR.
The Page component does not have any styles that we apply to it specifically. I have put in the scaffolding for Fela, in case we decide to put in styles during backbone-react migration.
If someone has objections to this I will revert to the original component, but it wont support Fela.


manatarms:
Just a reminder. Don't merge this till we know what to do with Fela components.

tajo:
please rebase

packages/cf-component-page/src/Page.js
```diff
@@ -1,17 +1,18 @@
...
-class Page extends React.Component {
-  render() {
-    return (
-      <article className="cf-page">
-        {this.props.children}
-      </article>
-    );
-  }
-}
+const styles = props => {
```
sejoker: `const styles = () => ({})`
is more concise.

packages/cf-component-page/src/PageHeader.js
```diff
@@ -1,20 +1,29 @@
...
-    return (
-      <header className="cf-page__header">
-        <h1 className="cf-page__title">{this.props.title}</h1>
-        {this.props.subtitle &&
-          <p className="cf-page__subtitle">{this.props.subtitle}</p>}
-      </header>
-    );
-  }
-}
+const styles = props => {
```
sejoker: I recommend using destructuring:

const styles = ({ theme }) => ({
    '& h1': {
      marginBottom: theme.marginBottom
     },
    '& p': {
       marginTop: theme.marginTop,
       fontSize: theme.fontSize,
      color: theme.defaultColor
     }
})

packages/cf-component-page/src/PageHeader.js
```diff
@@ -1,20 +1,29 @@
...
+      marginTop: theme.marginTop,
+      fontSize: theme.fontSize,
+      color: theme.defaultColor
+    }
+  };
+};
+const PageHeader = ({ title, subtitle, className }) => (
+  <header className={className}>
+    <h1 className={className}>{title}</h1>
+    {subtitle && <p className={className}>{subtitle}</p>}
```
sejoker: before your changes header and subtitle had different classNames, now you apply the same. It won't work like you expect.

packages/cf-component-page/test/__snapshots__/Page.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`should render 1`] = `
 <article
-  className="cf-page"
+  className=""
```
sejoker: this is an indication that something is broken, fela should always assign a className value.

tajo:
In case you need to assign multiple classNames to one React Component, you should split it into multiple React components. One component = 1 classname.

tajo:
> In case you need to assign multiple classNames to one React Component, you should split it into multiple React components. One component = 1 classname

Not true anymore. I've added `createComponentStyles` function into `cf-style-container` for this purpose. It now in this [PR](https://github.com/cloudflare/cf-ui/pull/133). See [the usage](https://github.com/cloudflare/cf-ui/pull/133/files#diff-fb2409e9f0fd8e3dc76ec94f8bc4f44eR2).

Anyway, your current solution doesn't work. You can't assign the same className to different nodes.

manatarms:
Should I close out this PR? A lot has changed since I made it.

wyuenho:
Also, please coordinate with https://github.com/cloudflare/cf-ui/pull/106

manatarms:
Closing this out because @jwineman is working with design to break up heading into distinct components.

PR #103, opened 2017-03-17
Update cf-component-label to use Fela

manatarms:
Creating this PR because I had done it. We can merge whenever we decide to.


manatarms:
As a reminder Do Not Merge.

tajo:
please rebase

manatarms:
@tajo This is good to go. You can merge this and publish into cf-ux when you're ready.

jwineman:
If @tajo approves lets merge and have him release.

I'm hesitant to release since I haven't been following the issues as closely recently.

tajo:
Question: Are these styles using our private styles?

packages/cf-component-label/example/basic/component.js
```diff
@@ -1,5 +1,8 @@
 import React from 'react';
-import Label from 'cf-component-label';
+import { applyTheme } from 'cf-style-container';
+import { Label as LabelUnstyled, LabelTheme } from 'cf-component-label';
+
+const Label = applyTheme(LabelUnstyled, LabelTheme);
```
tajo: We now export styled components by default. So this should go into `./src/index.js` and you should 

```js
export { Label, LabelUnstyled, LabelTheme }
```

manatarms:
Should I still make these changes?

wyuenho:
@manatarms Just fix this PR.

packages/cf-component-label/src/Label.js
```diff
@@ -1,16 +1,29 @@
...
-        className: 'cf-label cf-label--' + this.props.type
-      },
-      this.props.children
-    );
-  }
-}
+const styles = props => {
+  const theme = props.theme;
+  return {
+    display: theme.display,
```
manatarms: I used the public styles, this was before we decided to have a singular style sheet. I'll put in our private styles.

sejoker:
@wyuenho I disagree with migrating to ES6 format simply because of future requirements. The idea to use functional components is solely due to simpler mental model not because of speed benefits. IMO, the author who will need use of refs will convert to ES6 easily.

wyuenho:
>  the author who will need use of refs will convert to ES6 easily.

@sejoker I don't want to see any inconsistency. Needing refs is not a future requirement, it's a requirement now. This is probably something that should have gone into the discussions but didn't. We've talked about this about this time last year. No regression please, functional components don't buy us much now.

manatarms:
Please look at this so I can merge it. I've made all the changes as per the discussion #162

packages/cf-component-label/package.json
```diff
@@ -10,6 +10,10 @@
   "publishConfig": {
     "registry": "http://registry.npmjs.org/"
   },
+  "dependencies": {
+    "cf-style-container": "^1.0.0",
```
tajo: You don't need to change these manually. Lerna would handle this during the release.

packages/cf-component-label/src/Label.js
```diff
@@ -1,13 +1,35 @@
...
+    '-webkit-text-stroke': theme.webkitTextStroke,
+    whiteSpace: theme.whiteSpace,
+    backgroundColor: theme[`${props.type}BackgroundColor`]
+  };
+};
 
 class Label extends React.Component {
   render() {
-    return React.createElement(
-      this.props.tagName,
```
tajo: Why are we changing the API? We use `span`s everywhere?

packages/cf-component-label/src/LabelTheme.js
```diff
@@ -0,0 +1,27 @@
+import { rem } from 'polished';
+
+export default baseTheme => {
+    return {
+        borderRadius: baseTheme.borderRadius,
```
tajo: 4 spaces? should be 2

packages/cf-component-label/src/Label.js
manatarms: The reason is because I checked our codebase and the only implementation is using span. This also breaks our pattern of having a standardized component, a user could technically render anything using the tagName.

packages/cf-component-label/src/LabelTheme.js
manatarms: Ah man. Prettier defaults to 4.

jwineman: No code style code review comments!  Prettier is god.

packages/cf-component-label/src/Label.js
jwineman: in the same way `className` leaks style this was leaking DOM (lol?).  If we actually need to switch the underlying HTML tag we should expose it as an enumerated prop not a free form string.

packages/cf-component-label/src/LabelTheme.js
manatarms: Done :)

packages/cf-component-label/package.json
manatarms: Should I remove these? Or let them be?

packages/cf-component-label/src/Label.js
tajo: ok

packages/cf-component-label/package.json
tajo: You can leave it.

manatarms:
@wyuenho If this looks okay to you can we merge it?

PR #106, opened 2017-03-20
Update PageHeader to render subtitle as an h2

hturan:
This is a super-simple SEMVER-major PR to make `cf-component-page`'s `PageHeader` render it's `subtitle` prop as an `h2`.

I've run `npm run update-snapshot` and committed the changes for `cf-component-page`. Is this correct? I couldn't find any documentation about how to do this so just picked the most relevant `script` 😄

### DIFF
--- a/packages/cf-component-page/src/PageHeader.js
+++ b/packages/cf-component-page/src/PageHeader.js
@@ -6,7 +6,7 @@ class PageHeader extends React.Component {
       <header className="cf-page__header">
         <h1 className="cf-page__title">{this.props.title}</h1>
         {this.props.subtitle &&
-          <p className="cf-page__subtitle">{this.props.subtitle}</p>}
+          <h2 className="cf-page__subtitle">{this.props.subtitle}</h2>}
       </header>
     );
   }
--- a/packages/cf-component-page/test/__snapshots__/PageHeader.js.snap
+++ b/packages/cf-component-page/test/__snapshots__/PageHeader.js.snap
@@ -21,10 +21,10 @@ exports[`should render title/subtitle 1`] = `
   >
     Title
   </h1>
-  <p
+  <h2
     className="cf-page__subtitle"
   >
     Subtitle
-  </p>
+  </h2>
 </header>
 `;


sejoker:
LGTM

sejoker:
@hturan could lay out your reasoning behind the change.

hturan:
It looks like CI is failing on a completely separate suite; `cf-component-notifications`. Anyone know what's going on there?

@manatarms Ah ace, sorry for not checking! Feel free to make this change and close this when that lands 👍

@sejoker There are a variety of reasons, but mainly: if `title` is an `h1`, `subtitle` should semantically be an `h2` and also our current Backbone `page` implementation has the tag as an `h2`.

packages/cf-component-page/src/PageHeader.js
```diff
@@ -6,7 +6,7 @@ class PageHeader extends React.Component {
       <header className="cf-page__header">
         <h1 className="cf-page__title">{this.props.title}</h1>
         {this.props.subtitle &&
-          <p className="cf-page__subtitle">{this.props.subtitle}</p>}
+          <h2 className="cf-page__subtitle">{this.props.subtitle}</h2>}
```
jwineman: Why not `cf-component-heading`?

manatarms: I like the idea of using `cf-component-heading`, but it would introduce a dependency in `cf-component-page`

jwineman: I think we should avoid it whenever possible.  However we have instances where our components depend on each other.  `cf-component-card` has a dependency on `cf-component link` for example.

If we expect this `<h2>` to look and function the same as `cf-component-heading` we should introduce that dependency here since not doing so makes this use a customization that will limit its ability to get future updates we make to that component.  This has technically already happened since `cf-component-heading` has been converted to Fela but this component is still using hard coded CSS classes.

tajo:
@hturan 

> It looks like CI is failing on a completely separate suite; cf-component-notifications. Anyone know what's going on there?

Yea, it's a flaky test that sometimes time outs. :-/

tajo:
@hturan I reran travis. Works this time.

manatarms:
Don't merge this because I'm working on making PageHeader use Heading.

packages/cf-component-page/src/PageHeader.js
wyuenho: This is actually the right approach. There's only one kind of PageHeader, and the default styles in cf-component-heading are not contextual. `<Heading size={2}>` can mean very different things in different circumstances. There's no one size fits all approach here, as evident by the `<h1>` above.

koddsson:
Any updates on this? @manatarms it seems that #102 has been closed so I assume it's not blocking any more?

manatarms:
I think this is ok to merge.

koddsson:
@hturan Can we merge this or has this passed its last-sell-by date? 💩

hturan:
@koddsson I don't even know any more. Let's ship it!

koddsson:
:shipit: :shipit: :shipit:

koddsson:
Fuuuuuu, this is the wrong branch innit?

PR #108, opened 2017-03-20
Fix heading version in modal

tajo:
Lerna bumped heading's major version (fela version) and now cf-component-modal is failing in www-next because there is no fela yet. So I'm lowering the heading version to pre-fela version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "updated": "lerna updated --only-explicit-updates",
     "prettier": "node prettier.js",
     "clean": "lerna clean --yes && rimraf .jest",
-    "publish": "git pull --rebase && check-installed-dependencies && npm run build && lerna publish --only-explicit-updates",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "updated": "lerna updated --only-explicit-updates",
     "prettier": "node prettier.js",
     "clean": "lerna clean --yes && rimraf .jest",
-    "publish": "git pull --rebase && check-installed-dependencies && npm run build && lerna publish --only-explicit-updates",
+    "publish": "git pull --rebase && check-installed-dependencies && npm run build && lerna publish",
```
wyuenho: Revert this please. We don't want half a dozen packages with no code changes getting a version bump just because some util they all depend on get one.

jwineman: What was the decision on this?  Did we decide not to revert it?

sejoker: I didn't review this change initially, I agree with @wyuenho, let's be explicit in what we publish

PR #109, opened 2017-03-21
Export the http util methods explicitly so cf-util-http-file can cons…

wyuenho:
…ume them


sejoker:
@wyuenho @tajo when are we going to publish a new version to npm?

wyuenho:
Not yet. Actually @tajo is the only one who can publish now because npm stupidity.

Still trying to resurrect the cf-util-http tests and remove cf-pony.

PR #115, opened 2017-03-22
Refactor cf-component-text, cf-component-header, small bugfixes

sejoker:
according to feedback from @tajo:

* using plain object for applying styles
* exports components as `TextUnstyled`, `HeaderUnstyled` in tests to follow same style
* fix for text titlecase
* fix for text small size


sejoker:
@tajo are you ok with merging without second review?

manatarms:
Does this PR also include the things I listed in the JIRA?

sejoker:
@manatarms it's not, for testing the look and feel use cf-styleguide alongside cf-ux.

PR #116, opened 2017-03-22
Don't merge: recover of npm owner scripts

sejoker:
This reverts commit fac240c468ea189965ce13c7f007066fcfd4e6bc.

Due to issues with publishing to Cloudflare npm organization we may consider using old approach for the time being.

### DIFF
--- a/NPM_OWNERS
+++ b/NPM_OWNERS
@@ -1,0 +1,12 @@
+thejameskyle
+toekneestuck
+jwineman
+terinjokes
+triggity
+marksteyn
+wyuenho
+akenn
+thellimist
+
+koddsson
+jculvey
--- a/scripts/npm-owner-grant.sh
+++ b/scripts/npm-owner-grant.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+set -e
+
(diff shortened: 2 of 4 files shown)


jwineman:
Are the issues they don't have access? Give me a list of who needs access and I'll get it done today.

wyuenho:
It would help us a lot if you could make sure everyone is a maintainer on every cf-ui package on npm.

jwineman:
I'll bug @toekneestuck about this

wyuenho:
@jwineman @tajo @toekneestuck No need for this any more correct? Just add an npm user to and org account and done from now on?

tajo:
Let's wait with this. I don't think so we will be releasing any new packages in next few weeks and npm can fix it in meantime.

jwineman:
@wyuenho correct, I think.  The one caveat is whoever releases any new packages needs to make sure they add them to the org and not the personal account.

tajo:
I have admin rights to our npm org now, so if you need to publish a completely new cf-ui package, let me do it. Otherwise, we are fine.

PR #117, opened 2017-03-23
#67 replace superagent with fetch, revive util-http tests

wyuenho:
This PR replaces superagent with [fetch](https://developer.mozilla.org/en/docs/Web/API/Fetch_API). Because React already comes with `isomorphic-fetch`, this will allow us to get a smaller bundle size in the product.

cf-test-server is removed because jest.

The cf-util-http API is largely preserved. The only breaking change is the returned value from `request` is also changed to the promise returned from `fetch`. This is okay as practically no one uses the return value, and the only place in cf-ui has been removed.


hturan:
> React already comes with `isomorphic-fetch`

Really? I've never heard this before and can't find any evidence. Could you provide a link?

packages/cf-util-http/package.json
```diff
@@ -12,6 +12,10 @@
   },
   "dependencies": {
     "cf-util-logger": "^2.2.0",
-    "superagent": "^1.4.0"
-  }
+    "isomorphic-fetch": "^2.2.1"
+  },
+  "optionalDependencies": {
+    "es6-promise": "^4.1.0"
```
sejoker: why do we need es6-promise?

package.json
```diff
@@ -49,7 +51,9 @@
     "fela-plugin-validator": "^4.2.6",
     "file-loader": "^0.10.1",
     "glob": "^7.0.3",
+    "isomorphic-fetch": "^2.2.1",
```
sejoker: please update `yarn.lock`

package.json
```diff
@@ -27,7 +30,6 @@
     "babel-core": "^6.18.0",
     "babel-eslint": "^7.1.1",
     "babel-jest": "^19.0.0",
-    "babel-loader": "7.0.0-beta.1",
```
sejoker: nice catch

packages/cf-util-http-poll/src/httpPoll.js
```diff
@@ -48,12 +48,14 @@ export default function httpPoll(url, opts) {
   function onSuccess(res) {
     currentAbort = null;
     if (isCancelled) return;
-    if (!opts.isComplete(res)) {
-      intervalId = setTimeout(next, interval);
-    } else {
-      cleanup();
-      opts.onComplete(res);
-    }
+    opts.isComplete(res).then(completed => {
```
sejoker: what about .catch() for isComplete(res)

packages/cf-util-http-file/src/downloadFile.js
```diff
@@ -6,8 +6,10 @@ export default function downloadFile(url, filename, callback) {
     if (err) {
       callback(err);
     } else {
-      saveFile(filename, res.text);
-      callback(null, res);
+      res.text().then(body => {
```
sejoker: .catch()?

packages/cf-util-http/src/http.js
```diff
@@ -33,84 +25,80 @@ export function beforeSend(callback) {
...
+        logSuccess(logMessage);
+        if (callback) callback(undefined, res);
+      } else {
+        logError(logMessage);
+        if (callback) callback(res);
+      }
+
+      return res;
+    })
+    .catch(console.trace); // Unrecoverable errors
```
sejoker: console.error is more appropriate

packages/cf-util-http/test/http.js
```diff
@@ -1,94 +1,85 @@
...
         message: 'Hello World'
+      }),
+      {
+        headers: { 'Content-Type': 'application/json' },
+        status: 200
       }
     );
 
-    http.request('DELETE', '/posts/3', null, (err, res) => {
+    http.request('DELETE', '/posts/3', undefined, (err, res) => {
```
sejoker: what's the point of changing from null to undefined?

sejoker:
@hturan i'ts trivial to proof: https://github.com/facebook/react/blob/master/yarn.lock#L3137

hturan:
@sejoker Huh, part of `fbjs` (https://github.com/facebook/fbjs/blob/master/packages/fbjs/package.json#L64). I had no idea!

packages/cf-util-http/package.json
wyuenho: It's optional and doesn't get installed by default. Only needed for whoever uses this package on an environment that doesn't have a spec-compliant Promise, like node v4. It's a dependency of isomorphic-fetch.

packages/cf-util-http-poll/src/httpPoll.js
wyuenho: No need, `fetch` always return a resolved promise. The only exception is when you have a syntax or type error in the params to `fetch()`, which is unrecoverable and caught in `cf-util-http`.

packages/cf-util-http/src/http.js
wyuenho: This will show the stacktrace.

packages/cf-util-http/test/http.js
wyuenho: `null` is defined, `undefined` is not. Default argument values won't work if you passed in null. This is a slight API change but that's ok, no one uses `request()` directly.

tajo:
Do we want to export http.request?
Is there some reason why not to limit people only to get, post, delete... functions?

Other than, it's great! I'm really glad that the fakeServer is gone.

packages/cf-util-http/src/http.js
```diff
@@ -33,84 +25,136 @@ export function beforeSend(callback) {
...
+ *
+ * @see [Global Fetch on MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch)
+ * @see [WHATWG Fetch API Spec](https://fetch.spec.whatwg.org/)
  */
-export function request(method, url, opts, callback) {
-  opts = opts || {};
-
-  opts.method = method;
+export function request(method, url, opts = {}, callback) {
+  opts.method = method.toUpperCase();
```
sejoker: do we want a good default?
`opts.method = method ? method.toUpperCase() : 'GET';`

packages/cf-util-http/src/http.js
```diff
@@ -33,84 +25,136 @@ export function beforeSend(callback) {
...
-    req.query(opts.parameters);
-  }
-
-  if (opts.headers) {
-    req.set(opts.headers);
-  }
-
-  if (opts.body) {
-    req.send(opts.body);
+    if (url.indexOf('?') === -1) {
```
sejoker: `if (url && url.indexOf())`

packages/cf-util-http/src/http.js
```diff
@@ -33,84 +25,136 @@ export function beforeSend(callback) {
...
-    req.abort();
-  };
+      return response.text().then(text => {
+        if (callback) {
+          callback(wrapResponse(headers, status, text, text, response));
+        }
+        return response;
+      });
+    })
+    .catch(console.trace); // Unrecoverable errors
```
sejoker: I think callback should be always be called event if fetch failed due to network error
```
.catch(error => {
  if (callback) {
          callback(error);
         logError(error);
        }
})
```

wyuenho:
@tajo Request has its uses, in case you want to make a HEAD request or something.

packages/cf-util-http/src/http.js
wyuenho: No. This method is meant for the case you know what you are doing but there's no helper to help you. If you want to send a GET request, just use `http.get()`

packages/cf-util-http/src/http.js
wyuenho: Just let it die in this case. You can't do anything without a URL.

packages/cf-util-http/src/http.js
```diff
@@ -33,84 +25,136 @@ export function beforeSend(callback) {
...
+  }
+  return hash;
+}
+
+function wrapResponse(headers, status, body, text, response) {
+  return {
+    headers,
+    status,
+    body: headers['content-type'] === 'application/json'
+      ? JSON.parse(text)
```
sejoker: I would recommend `response.json().then()`
JSON.parse could throw, we don't handle this

packages/cf-util-http/src/http.js
wyuenho: The promise is returned, so the caller can catch it there and do the appropriate thing. The error is unrecoverable as this point by fetch, I just we should just log it and let the caller handle it, or ignored if desired. The only `error` the caller can handle is anything not 2xx. The callback is called with an error in a resolved promise state to preserve the existing API.

packages/cf-util-http/src/http.js
wyuenho: I'll put a try catch in request(), the promise should never be rejected here.

jwineman:
Thanks for doing this Jimmy

PR #119, opened 2017-03-23
Open source our private styles

tajo:
We are now opening our private internal styles.

- move internal (cf-ux) themes for cf-component-button, cf-component-text and cf-component-heading to cf-ui
- cf-ux/cf-style-const is going to be deleted, constants moved to cf-ui/cf-style-const
- the rest of components (not transitioned yet) use our internal postcss styles now (this change is visible in a different PR in cf-styleguide)
- rebuild of https://cloudflare.github.io/cf-ui/ (`/docs` folder)

**This is big since we don't have to keep two sets of styles (public vs private) and it makes our life easier.** 🍾


tajo:
@sejoker https://github.com/blog/2233-publish-your-project-documentation-with-github-pages `/docs` is how we publish our github page, so it needs to be a part of this repo

If you want to remove some of these fonts it needs to be done in cf-www-next's styles (it's the source of truth for styleguide).

koddsson:
>cf-ux/cf-style-const is going to be deleted, constants moved to cf-ui/cf-style-const

What's `cf-style-const`? Can it live in `cf-ui`?

tajo:
@koddsson it's a global theme, an object with variables like colors and fonts. It's passed to context of our apps. This global theme can be overwritten by local component themes, local themes can also create a new values based on global values. In some ways it replaces cascading.

Since we don't have "private only" styles, there's need only for one global theme and we will keep it here in cf-ui.

PR #120, opened 2017-03-24
Add example how to start with cf-ui and Fela

tajo:
**Since cf-ui doesn't work without Fela it's time to add a complete example how to setup the cf-ui+Fela project.**

I already did it for www-next. We need @jwineman to do the same for partners and @nkcmr for ninja panel. Do we use cf-ui anywhere else?

When it's done we can use the latest cf-ui components everywhere without compromising our workflow no matter how far the transition is.


thejameskyle:
Seems like you should abstract all of this setup into a single function you can import and call with some arguments.

tajo:
@thejameskyle Yea, I was thinking about that. https://github.com/cloudflare/cf-ui/issues/123

jwineman:
Is there any benefit to me doing this to partners _before_ we finish the migration?  I was planning on doing it after we ironed everything out.

tajo:
@jwineman yea, testing... anyway, I'm working on cf-style-provider (https://github.com/cloudflare/cf-ui/issues/123) now, so use that when published instead of this manual setup.

PR #121, opened 2017-03-24
Allow nodes as content in cf-component-callout

koddsson:
I also changed title to be not required.


tajo:
Add reviewers (you need at least two approvals). :)

koddsson:
@tajo I opened it up in a hurry :)

PR #122, opened 2017-03-24
Unpin cf-component-heading in cf-component-modal

tajo:
There is no reason why old components couldn't consume new Fela components if Fela is supported in www-next now. It's critical to not pin cross dependencies in order to make lerna and our workflow happy.

Heading is bumped to v3 (fela version) in cf-component-modal, so the theme needs to be applied. We still major bump cf-component-modal so it doesn't break anything.


wyuenho:
Okay, just make sure you bump the major version of cf-component-modal when you release

PR #124, opened 2017-03-26
Example: Upgrade deps and add React

tajo:
Somehow I forgot to add React as a dependency. 🙈

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -14,7 +14,7 @@
     "babel-preset-es2015": "^6.24.0",
     "babel-preset-react": "^6.23.0",
     "cf-component-button": "^4.2.0",
-    "cf-style-const": "^0.2.0",
+    "cf-style-const": "^0.3.0",
     "cf-style-container": "^0.2.0",
     "fela": "^4.3.2",
     "fela-font-renderer": "^4.3.2",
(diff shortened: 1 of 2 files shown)


example/package.json
```diff
@@ -14,14 +14,15 @@
...
-    "cf-style-const": "^0.2.0",
+    "cf-style-const": "^0.3.0",
     "cf-style-container": "^0.2.0",
     "fela": "^4.3.2",
     "fela-font-renderer": "^4.3.2",
     "fela-plugin-fallback-value": "^4.3.2",
     "fela-plugin-lvha": "^4.3.2",
     "fela-plugin-prefixer": "^4.3.2",
     "fela-plugin-unit": "^4.3.2",
+    "react": "^15.4.2",
```
jwineman: Any reason not to support the other versions?
`"react": "^0.14.0 || ^15.0.0-0 || ^16.0.0-0"`

tajo: This type of range makes sense for peerDeps but not here. This (example) is not a lib but the whole app.

tajo:
I'm closing this because it's fixed in https://github.com/cloudflare/cf-ui/pull/128 as well.

PR #125, opened 2017-03-27
card: Keep a local state of drawers if no props

koddsson:
Whenever you're using the `CardDrawers` sub-component of `Card` you need
to add some keys to your state and handle all the events yourself. This
is fine if you want to control the flow very tightly but the most common
case is just to switch to the drawers selected.

This patch changes the `CardDrawers` sub-component so it still takes in
the props and those will control the component but falls back to the
local state if the props `active` and `onClick` are not present.

Not sure if there are more tests to be written or if the snapshots are enough.

### DIFF
--- a/packages/cf-component-card/src/CardDrawers.js
+++ b/packages/cf-component-card/src/CardDrawers.js
@@ -26,7 +26,9 @@ class CardDrawers extends React.Component {
   }
 
   render() {
-    const active = this.props.active || this.state.active;
+    const active = this.props.hasOwnProperty('active')
+      ? this.props.active
+      : this.state.active;
 
     const links = [];
     const drawers = [];
--- a/packages/cf-component-card/src/CardDrawers.js
+++ b/packages/cf-component-card/src/CardDrawers.js
@@ -9,15 +9,30 @@ let UNIQUE_ID = 0;
 class CardDrawers extends React.Component {
   constructor(props) {
     super(props);
+
     this._cardId = UNIQUE_ID++;
+
+    this.state = {
+      active: null
+    };
+  }
+
+  handleLinkClick(id) {
+    this.props.onClick && this.props.onClick();
+
+    this.setState(state => ({
+      active: state.active === id ? null : id
+    }));
   }
 
   render() {
+    const active = this.props.active || this.state.active;
+
     const links = [];
     const drawers = [];
 
     this.props.drawers.forEach(drawer => {
-      const isActive = drawer.id === this.props.active;
+      const isActive = drawer.id === active;
       const id = `card-${this._cardId}-drawer-${drawer.id}`;
 
       links.push(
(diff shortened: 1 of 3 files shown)


packages/cf-component-card/src/CardDrawers.js
```diff
@@ -9,23 +9,38 @@ let UNIQUE_ID = 0;
...
+  handleLinkClick(id) {
+    this.props.onClick && this.props.onClick();
+
+    this.setState(state => ({
+      active: state.active === id ? null : id
+    }));
   }
 
   render() {
+    const active = this.props.active || this.state.active;
```
tajo: Are you sure it works the way you expect? When you click on the` CardToolbarLink` it will set `state.active = drawer.id` and even if your `active` prop is set explicitly to null, the state will take over.

hturan: Yep, haha, that. We just had a discussion 😄

packages/cf-component-card/src/CardDrawers.js
```diff
@@ -9,23 +9,40 @@ let UNIQUE_ID = 0;
...
+  handleLinkClick(id) {
+    this.props.onClick && this.props.onClick();
+
+    this.setState(state => ({
+      active: state.active === id ? null : id
+    }));
   }
 
   render() {
+    const active = this.props.hasOwnProperty('active')
```
tajo:
```jsx
<CardDrawers active={undefined}>
```
results in

```js
this.props.hasOwnProperty('active') === true
```

right? It's probably better to test if `this.props.active` is `undefined` instead.

koddsson: Yeah. I was thinking `undefined` and `null` were kind of interchangable here. If you pass in `null` or `undefined` you want no drawer to be open?

tajo: When you pass `undefined` I would expect that

```js
active = this.state.active;
```

so it behaves the same as in a case the prop `active` was left out completely.

When `null`

```js
active = this.props.active;
```

But here it will assign `this.props.active` in both cases. `undefined` means to me the prop doesn't exist and the component should use its own state. `null` means the prop exist but we want no drawer to be open.

Your call. If this works for you I'm fine.

koddsson: If it's my call, I'm gonna go with this because I need it :p

hturan: Yeah, originally I figured that if you're explicitly setting a prop then you'd like to affect the outcome, but checking against `undefined` is simpler.

hturan:
@tajo, is it correct that this is going into `master` rather than `legacy`? I just had one of my PRs rebased against `legacy`. What's the procedure here? If so, can we update CONTRIBUTING to be clearer?

tajo:
@hturan **Keep using master**. When to use legacy?

- your app (cf-www-next) is using lower major version than master has
- you don't want to update to the latest major version

Then, you have to create a PR against both legacy and master and release both branches. So legacy is for cases when you are using and updating older major versions.

Issue #123, opened 2017-03-25
cf-style-provider


tajo:
We should abstract Fela support into its own package since we now do very similar setup in cf-ux (for testing), cf-ui (for testing), cf-styleguide, cf-www-next, NP and partners. It would automatically also add our global theme (cf-style-const) into the context.

**New simplified setup**

```jsx
import { Provider } from 'cf-style-provider';
import { Button, ButtonTheme } from '@clouflare/cf-component-button';

const fontNode = document.getElementById('font-stylesheet');
const cssNode = document.getElementById('stylesheet');

render(
  <Provider fontNode={fontNode} cssNode={cssNode}>
      <Button type="primary" onClick={() => console.log('clicked')}>
        Primary Button
      </Button>
  </Provider>
);
``` 

compare it with the example in README. I think that 3rd party consumers of cf-ui still might want to rather do more manual setup since they don't want to be locked in into our global constants and set of Fela plugins but useful for us.

PR #128, opened 2017-03-27
Add cf-style-provider

tajo:
Explained here https://github.com/cloudflare/cf-ui/issues/123.

It's already dogfooded by the `/example` and tests (felaTestContext). Once published, I will introduce it into cf-styleguide, cf-ux and cf-www-next as well.


packages/cf-style-provider/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  return createFelaRenderer({
+    plugins,
+    enhancers,
+    selectorPrefix: [usedOpts.selectorPrefix]
+  });
+};
+
+export const createProvider = opts => {
```
thejameskyle: Why is this a HOC?

tajo: Because you need to configure the [renderer](https://github.com/rofrischmann/fela/blob/master/docs/basics/Renderer.md). It doesn't seem like something that should be done in `render()` ?

packages/cf-style-provider/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+  const renderer = createRenderer(opts);
+  const cssNode = opts && opts.cssNode ? opts.cssNode : undefined;
+  return class StyleProvider extends Component {
```
sejoker: could use functional component instead

packages/cf-style-container/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "css-color-function": "^1.3.0",
+    "fela": "^4.3.2",
```
sejoker: what's the difference between `cf-style-container` and `cf-style-provider`. To me they could be merged into single component.

sejoker: please lock dependencies with yarn.lock. We don't want failed CI builds every time fela upgrades

packages/cf-style-provider/package.json
```diff
@@ -0,0 +1,31 @@
+{
```
sejoker: please lock dependencies with yarn.lock. We don't want failed CI builds every time fela upgrades

packages/cf-style-provider/README.md
```diff
@@ -0,0 +1,51 @@
...
+```jsx
+import ReactDOMServer from 'react-dom/server';
+import { Provider } from 'react-fela';
+import { createRenderer } from 'cf-style-provider';
+
+// default values
+const renderer = createRenderer({
+  cssNode: null, // where to render CSS styles (DOM node)
+  fontNode: null, // where to render CSS styles related to fonts (DOM node)
+  dev: false, // in dev mode it applies more Fela plugins (validator, ...)
```
sejoker: beautifier and monolithic as well. It would be useful explaining why we need them.

packages/cf-style-container/package.json
tajo: `cf-style-container` is a set of helpers that are used **by Fela components**. `cf-style-provider` is used **by consumers of cf-ui**. We definitely don't want to merge them because `cf-style-provider` has a lot of Fela dependencies (plugins, enhancers etc) that have no use for components. They don't overlap at all.

packages/cf-style-provider/package.json
tajo: I would like to but we depend on lerna here. The yarn support is [experimental](https://github.com/lerna/lerna/issues/371), let's wait when it's more stable.

> We don't want failed CI builds every time fela upgrades

It should not be *every* time.

packages/cf-style-provider/README.md
tajo: done

packages/cf-style-provider/src/index.js
tajo: done

packages/cf-style-provider/package.json
sejoker: I could test [yarn support|https://github.com/lerna/lerna/releases/tag/v2.0.0-beta.38]. It allows install with yarn and nothing else, this is exactly what we need.

packages/cf-style-provider/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "jsnext:main": "es/index.js",
+  "author": "Vojtech Miksu <vojtech@cloudflare.com>",
+  "license": "BSD-3-Clause",
+  "publishConfig": {
+    "registry": "http://registry.npmjs.org/"
+  },
+  "dependencies": {
+    "cf-style-const": "^0.3.0",
+    "cf-style-container": "^0.2.0",
+    "fela": "^4.3.2",
```
sejoker: An idea: could we declare all fela dendencies as optional and keep them in the root package.json?

packages/cf-style-provider/package.json
tajo: Ok, but let's do in a different PR. There should be yarn.lock in every package. Update travis.yml. Test that it works for private packages as well (cf-ux).

sejoker: @tajo agree.

packages/cf-style-provider/package.json
tajo: Root package.json? Do you mean cf-ui/package.json? I don't think so. 

Pretty much the whole point of cf-style-provider is to shield you from setting up Fela if you want to consume cf-ui components and provide you with all necessary dependencies.

packages/cf-style-provider/src/index.js
thejameskyle: I don't know why, they are both done statically. It's not going to run multiple times.

tajo:
I removed HOC and simplified usage of `StyleProvider` as @thejameskyle suggested. @jwineman and @sejoker please take a quick look again.

jwineman:
LGTM

sejoker:
LGTM

Issue #50, opened 2017-01-04
Remove cf-util-text and deprecate the package on npm


wyuenho:
This was only introduced for copyable-textarea at the 11th hour, this dependency has been removed.

UI library code should only provide a way to customize text as opposed to mandating yet another half done gettext style library.

PR #130, opened 2017-03-28
Issue #50 remove useless packages

wyuenho:
No package depends on cf-pony anymore, and cf-util-text was only introduced in the 11th hour to deal with i18n. I18n should not be cf-ui's responsibility, text should already be translated when they are passed to React components for rendering.


jwineman:
I think there is value in keeping `cf-component-text`.  Originally the argument against it was the fact that you can't control inherited styles from leaking into the component around fonts right?  Isn't this solved with Fela?

If we remove it how will we consistently align/style/size text?

sejoker:
@wyuenho could you explain why we are removing `wrapping-external-apis.md`? Is it not a right approach to follow? I agree with forcing minimal API surface area. Instead of removing cf-pony, should we discuss how we could improve the usage, what was the painpoints we have?

jwineman:
I agree with not removing the coding to the interface discussion, especially since we just exercised an example of it switching `cf-util-http` from super agent to fetch.

wyuenho:
@jwineman I'm not remove cf-component-text, I'm removing cf-util-text.
@sejoker Introducing yet another layer of abstraction is not going to solve the problem of breaking changes in external APIs. If that was a valid argument, the first thing we should wrap is React, and then everything else, trust no one. One of the major reason that was written was to provide justification for cf-pony. At some point I thought, all of those APIs are written dead in various industry standards and specs, if they break, all hell breaks loose. Sometimes, we just have to have a little faith in the people that write libraries for us. If they don't keep the APIs stable, maybe the solution is to not use them at all, instead of writing a facet to smooth over other people's faults.

jwineman:
Ah okay, I'm fine removing `cf-util-text`.

I think wrapping React in an interface is overkill (which is why we didn't do it) but I think programming to the interface is an important concept and our use of it has been incredibly helpful.

In general I also don't like removing discussions.  If we change our mind about something we should "deprecate" them or something and explain why we went in a new direction.  The purpose of the discussions is to explain our thought process as well document best practices for using the library.

wyuenho:
There's no "our" in the discussion, neither was it a discussion. Discussions involve some back and forth, those documents were essentially decisions.
This is the third time we remove documents from it because they don't make sense anymore. The thing is, as you use the library and look at the code, the tradeoffs will be apparent. If it's not, there's something wrong with it, so you file an issue, and the rationale will be on Github. If you look at the remaining documents, anything related to redux show be gone when we rip out the redux stuff from the builders, the part about lerna and monorepo may also change, depends on what we want to do with lerna. What's left is essentially common sense.

jwineman:
Okay fine, but in this particular case we're still coding to the interface with cf-util-http so why do you want to remove it?

wyuenho:
Wrapping an API allows us to easily swap out implementation, not prevent breakage. It also presupposes that the wrapper's API won't change. This cannot be the case since cf-util-http's API is already broken. We are the only people we support, and updating a dozen call sites in internal projects is trivial. The premise in that document is common sense, the rationale is not.

In general, I'd very much like not to have those documents at all. They make people hesitant to evolve the code base. I'd be fine to have a set of documents to justify the decisions if cf-ui was "done". It's far from it.

jwineman:
I guess if you feel strongly about it thats fine.  It seems like this is more about removing all discussions and less about removing this particular discussion though.

I think the difference between you and me is how strict we view the the discussions.  The only one I was super passionate about was not allowing `className`.  The rest of them I view as best practices rather than hard rules.

Taking cf-util-http as an example specifically I don't think it means we'll never break the API, but more that it forces us to think about whether or not we want to.  If we don't have the interface we're going to break the implementation every time.  If we do have it then breaking it is an option, not a default.

PR #132, opened 2017-03-28
Emulate superagent's automatic encoding of JSON body

wyuenho:
This is rather urgent as some internal code just assumes the body will be automatically encoded

### DIFF
--- a/packages/cf-util-http/src/http.js
+++ b/packages/cf-util-http/src/http.js
@@ -69,6 +69,8 @@ function wrapResponse(headers, status, body, text, response) {
   };
 }
 
+const BODYLESS_METHODS = ['GET', 'HEAD'];
+
 /**
  * Perform an http request. This method uses WHATWG's fetch API and returns a
  * promise that **resolves on any HTTP status**, and rejects on anything else,
@@ -112,6 +114,8 @@ export function request(method, url, opts, callback) {
     opts = {};
   }
 
+  opts = Object.assign({}, opts);
+
   opts.method = method.toUpperCase();
   opts.url = url;
   opts.callback = callback;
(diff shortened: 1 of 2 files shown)


wyuenho:
@jculvey @tajo @manatarms @jwineman I just simplified the code a lot, can you re-review again 🙏🙏🙏🙏🙏🙏🙏🙏

packages/cf-util-http/src/http.js
```diff
@@ -123,14 +127,21 @@ export function request(method, url, opts, callback) {
   url = opts.url;
   callback = opts.callback;
 
+  // Emulate superagent's ability to automatically encode JSON body and set the
+  // Content-Type
+  if (BODYLESS_METHODS.every(m => method.toUpperCase() !== m) && opts.body) {
+    opts.body = JSON.stringify(opts.body);
+    opts.headers = new Headers(opts.headers || {});
+    opts.headers.set('Content-Type', 'application/json');
```
jwineman: Does this replace the `Content-Type` header if it already exists or add a second one?  I can't find docs for `.set()` they all link to `Set()`(Data structure)

wyuenho: https://developer.mozilla.org/en-US/docs/Web/API/Headers/set

jwineman: Doh.  Thanks @wyuenho 
https://developer.mozilla.org/en-US/docs/Web/API/Headers/set

wyuenho: I'll make it conditional

packages/cf-util-http/src/http.js
```diff
@@ -112,6 +114,8 @@ export function request(method, url, opts, callback) {
     opts = {};
   }
 
+  opts = Object.assign({}, opts);
```
sejoker: love this change.

PR #133, opened 2017-03-29
Convert cf-component-form to Fela

tajo:
This is ready to go!

### DIFF
--- a/packages/cf-component-form/src/Form.js
+++ b/packages/cf-component-form/src/Form.js
@@ -1,20 +1,33 @@
 import React, { PropTypes } from 'react';
+import { createComponent } from 'cf-style-container';
 
-class Form extends React.Component {
-  render() {
-    let className = 'cf-form cf-form--' + this.props.layout;
-    return (
-      <form className={className} onSubmit={this.props.onSubmit}>
-        {this.props.children}
-      </form>
-    );
-  }
-}
+const styles = ({ theme }) => ({
+  background: theme.background,
+  border: theme.border,
+  boxShadow: theme.boxShadow
+});
+
+const addLayoutProp = (children, layout) =>
+  React.Children.map(children, (child, index) => {
+    if (React.isValidElement(child)) {
+      return React.cloneElement(child, { layout });
+    }
+    return child;
+  });
+
+const Form = ({ children, layout, onSubmit, className }) => (
+  <form className={className} onSubmit={onSubmit}>
+    {addLayoutProp(children, layout)}
+  </form>
+);
 
 Form.propTypes = {
   layout: PropTypes.oneOf(['horizontal', 'vertical']).isRequired,
+  className: PropTypes.string,
   onSubmit: PropTypes.func.isRequired,
   children: PropTypes.node
 };
 
-export default Form;
+Form.displayName = 'Form';
+
+export default createComponent(styles, Form);
--- a/packages/cf-component-form/src/FormFieldset.js
+++ b/packages/cf-component-form/src/FormFieldset.js
@@ -44,7 +44,6 @@ const contentStyles = ({ theme, layout }) => ({
 
 const FormFieldset = ({ legend, children, styles }) => (
   <fieldset className={styles.mainStyles}>
-    {' '}
     <legend className={styles.legendStyles}>
       {legend}
     </legend>
--- a/packages/cf-component-form/src/FormFieldset.js
+++ b/packages/cf-component-form/src/FormFieldset.js
@@ -1,5 +1,5 @@
 import React, { PropTypes } from 'react';
-import { createComponentStyles, combineRules } from 'cf-style-container';
+import { createComponentStyles } from 'cf-style-container';
 import { clearFix } from 'polished';
 
 const mainStyles = ({ theme }) => ({
--- a/packages/cf-component-form/src/FormFieldset.js
+++ b/packages/cf-component-form/src/FormFieldset.js
@@ -1,23 +1,67 @@
 import React, { PropTypes } from 'react';
+import { createComponentStyles, combineRules } from 'cf-style-container';
+import { clearFix } from 'polished';
 
-class FormFieldset extends React.Component {
-  render() {
-    return (
-      <fieldset className="cf-form__fieldset">
-        <legend className="cf-form__fieldset_legend">
-          {this.props.legend}
-        </legend>
-        <div className="cf-form__fieldset_content">
-          {this.props.children}
-        </div>
-      </fieldset>
-    );
-  }
-}
+const mainStyles = ({ theme }) => ({
+  margin: theme.main.margin,
+  padding: theme.main.padding,
+  borderTop: theme.main.borderTop,
+  borderLeft: theme.main.borderLeft,
+  borderRight: theme.main.borderRight,
+  borderBottom: theme.main.borderBottom,
+  ...clearFix()
+});
+
+const legendStyles = ({ theme, layout }) => ({
+  padding: theme.legend.padding,
+  marginBottom: theme.legend.marginBottom,
+  borderBottom: theme.legend.borderBottom,
+  '@media (min-width: 720px)': layout === 'horizontal'
+    ? {
+        width: '30%',
+        float: 'left',
+        padding: '1.7em 1rem',
+        color: theme.colorGrayDark,
+        textAlign: 'right'
+      }
+    : {}
+});
+
+const contentStyles = ({ theme, layout }) => ({
+  padding: theme.content.padding,
+  border: theme.content.border,
+  borderTopWidth: theme.content.borderTopWidth,
+  background: theme.content.background,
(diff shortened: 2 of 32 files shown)


packages/cf-component-form/src/Form.js
```diff
@@ -1,20 +1,33 @@
...
-  }
-}
+const styles = ({ theme }) => ({
+  background: theme.background,
+  border: theme.border,
+  boxShadow: theme.boxShadow
+});
+
+const addLayoutProp = (children, layout) =>
+  React.Children.map(children, (child, index) => {
```
sejoker: nice approach

packages/cf-component-form/src/FormFieldset.js
```diff
@@ -1,23 +1,67 @@
...
-  }
-}
+const mainStyles = ({ theme }) => ({
+  margin: theme.main.margin,
+  padding: theme.main.padding,
+  borderTop: theme.main.borderTop,
+  borderLeft: theme.main.borderLeft,
+  borderRight: theme.main.borderRight,
+  borderBottom: theme.main.borderBottom,
+  ...clearFix()
```
sejoker: 👍

packages/cf-component-form/src/FormFieldset.js
```diff
@@ -1,23 +1,67 @@
...
+});
+
+const legendStyles = ({ theme, layout }) => ({
+  padding: theme.legend.padding,
+  marginBottom: theme.legend.marginBottom,
+  borderBottom: theme.legend.borderBottom,
+  '@media (min-width: 720px)': layout === 'horizontal'
```
sejoker: I like how flexible we could express styling

packages/cf-component-form/src/FormHeader.js
```diff
@@ -1,19 +1,23 @@
...
-  }
-}
+const styles = ({ theme }) => ({
+  padding: theme.padding,
+  borderBottom: theme.borderBottom
+});
+
+const FormHeader = ({ className, title }) => (
+  <div className={className}>
+    <Heading size={4}>
```
sejoker: it was h3 before, why use h4 now?

packages/cf-style-container/src/index.js
```diff
@@ -37,4 +39,24 @@ const applyTheme = (ComponentToWrap, theme) => {
   return ThemedComponent;
 };
 
-export { createComponent, applyTheme, ThemeProvider, color };
+const createComponentStyles = (styleFunctions, component) => {
```
sejoker: please add a description about use case for this HOC.

packages/cf-component-form/src/FormHeader.js
tajo: Because it's more similar to the old version where we remove margin when used inside of form (we should not do stuff like this) but our headings come with fixed margin. We most likely make more style changes pretty soon since we don't really use cf-component-form as we should.

packages/cf-style-container/src/index.js
tajo: Done. I've documented all `cf-style-container` functions and added a test for `createComponentStyles`.

packages/cf-style-container/src/index.js
```diff
@@ -37,4 +39,24 @@ const applyTheme = (ComponentToWrap, theme) => {
...
+
+export {
+  createComponent,
+  applyTheme,
+  ThemeProvider,
+  color,
+  connect,
+  combineRules,
+  createComponentStyles
+};
```
wyuenho: I still don't know why this package is called cf-style-container? It doesn't contain any styles. This looks like cf-util-styles to me.

packages/cf-component-form/src/Form.js
wyuenho: NVM, I grok this now.

packages/cf-style-container/README.md
```diff
@@ -2,10 +2,105 @@
...
 
 ## Installation
 
 ```sh
 $ npm install cf-style-container
 ```
+
+### Aliased functions from fela and react-fela
+
+We proxy/alias some useful functions from fela without changing their behaviour. See the original documentation for more details. We wrap all Fela APIs so we can eventually switch Fela to a different CSS in JS lib if ever needed.
```
tajo: Most of them share the same camelCase CSS in JS styles, so you can use any renderer if there's better. Btw, that's why projects like polished can be shared. It's still nice to have one `cf-style-container` toolbox.

packages/cf-component-form/src/FormTheme.js
```diff
@@ -0,0 +1,10 @@
+import { checkBaseTheme } from 'cf-style-const';
+
+export default baseTheme => {
+  checkBaseTheme(baseTheme, 'FormTheme');
+  return {
+    background: 'white',
+    border: `1px solid ${baseTheme.colorGrayLight}`,
```
tajo: Not sure what you mean by this.

packages/cf-component-form/src/FormLabel.js
```diff
@@ -1,24 +1,23 @@
...
-    }
-
-    return (
-      <label className={className}>
-        {this.props.children}
-      </label>
-    );
-  }
-}
+const FormLabel = ({ children, className }) => (
```
tajo: > We'll love to have a ref when interoping with other DOM manipulation libs

That sounds like something we definitely want to avoid?

> you'd think this may be faster because of some old blog post and some ancient twitter message, but it actually isn't.

It will be with Fiber.

packages/cf-style-container/src/index.js
tajo: We wanted to have all style related stuff under cf-style.

packages/cf-component-form/src/FormHeader.js
tajo: Done. I still used h3 since it matches the old styles. We will redesign all forms pretty soon anyway (more info coming soon).

packages/cf-component-form/src/FormLabel.js
tajo: Also, I like them because it's less code to write and it promotes not using internal state.

packages/cf-component-form/src/FormFieldset.js
```diff
@@ -1,23 +1,67 @@
...
+        float: 'left',
+        borderTopWidth: 0,
+        borderLeftWidth: 0
+      }
+    : {}
+});
+
+const FormFieldset = ({ legend, children, styles }) => (
+  <fieldset className={styles.mainStyles}>
+    {' '}
```
tajo: Fixed.

packages/cf-component-form/src/FormLabel.js
wyuenho: We'd like to avoid using refs, but can't. Refs also isn't just something used by DOM manip libs too, tho that's a big issue because tooltip and such. Let's wait til fiber.

tajo: Well, if that's everyone's opinion you should open a new PR and revert it everywhere since we use functional components within all new cf-ui codebase. I don't like it. If we really need refs, we can do it on per instance basis.

packages/cf-style-container/src/index.js
wyuenho: Then why is it called container?

packages/cf-component-form/src/FormTheme.js
wyuenho: Something like `expandShorthand({border: ''}) => {borderTop, borderRight...}`

packages/cf-style-container/README.md
wyuenho: Polished is just a bunch of "css preprocessor mixins" for CSS-in-JS solutions, basically, just a bunch of simple, basically pure functions. Fela has a lot of behavior that can't be expressed by these simple signatures. Replacing things is never really easy. Back in the days, people thought having an ORM layer would make switch the RDBMS easy, I lol'ed. Anyway, I digress, it's not important.

packages/cf-component-form/src/FormTheme.js
tajo: We should call this atomization rather than DCE. This would be even better:

```js
...expandBorder('1px solid black')
```

output

```js
borderTopColor: 'black',
borderLeftColor: 'black',
borderBottomColor: 'black',
borderRightColor: 'black',
borderTopColor: 'black',
borderLeftStyle: 'solid',
borderBottomStyle: 'solid',
borderRightStyle: 'solid',
borderTopWidth: '1px',
borderLeftWidth: '1px',
borderBottomWidth: '1px',
borderRightWidth: '1px'
```

packages/cf-style-container/src/index.js
tajo: because naming is hard, especially at the beginning when you don't exactly know what will go inside, complain to @jwineman, his idea

PR #134, opened 2017-03-29
Emulate superagent's ability to filter out headers with null or undefined values

wyuenho:
This should be the last fix for cf-util-http where people were taking advantage of superagent's ability to filter out null headers and now have the their APIs broken by fetch.

### DIFF
--- a/packages/cf-util-http/test/http.js
+++ b/packages/cf-util-http/test/http.js
@@ -104,11 +104,12 @@ describe('request', () => {
     http.request(
       'GET',
       '/somewhere',
-      { headers: { foo: null, bar: 'bar' } },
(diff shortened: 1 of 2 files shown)


packages/cf-util-http/test/http.js
```diff
@@ -82,6 +82,33 @@ describe('request', () => {
...
+    );
+    expect(fetch.mock.calls[0][1].body).toBe('hello world');
+    done();
+  });
+
+  test('should ignore null header values', done => {
+    http.request(
+      'GET',
+      '/somewhere',
+      { headers: { foo: null, bar: 'bar' } },
```
prestonp: Add `baz: undefined` and assert falsy for good measure?

PR #135, opened 2017-03-29
cf-component-pagination migration to Fela

sejoker:


jwineman:
Looks like `cf-builder-pagination` tests are failing

packages/cf-component-pagination/example/basic/component.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Pagination, PaginationItem } from 'cf-component-pagination';
 import Icon from 'cf-component-icon';
+import { applyTheme } from 'cf-style-container';
```
tajo: unused

packages/cf-component-pagination/src/Pagination.js
```diff
@@ -1,31 +1,53 @@
...
   info: PropTypes.string,
   children: PropTypes.node
 };
 
-export default Pagination;
+const PaginationUnstyled = createComponent(styles, Pagination);
+
+export { PaginationUnstyled };
+
+export default applyTheme(PaginationUnstyled, PaginationTheme);
```
tajo: To preserve the consistency with my code, can you do these exports in index.js instead?

packages/cf-component-pagination/src/PaginationItem.js
```diff
@@ -68,4 +92,8 @@ PaginationItem.propTypes = {
   children: PropTypes.node
 };
 
-export default PaginationItem;
+const PaginationItemUnstyled = createComponent(styles, PaginationItem);
+
+export { PaginationItemUnstyled };
+
+export default applyTheme(PaginationItemUnstyled, PaginationItemTheme);
```
tajo: please, do these exports in index.js

packages/cf-component-pagination/src/PaginationLink.js
```diff
@@ -0,0 +1,70 @@
...
+  onClick: PropTypes.func.isRequired,
+  clickable: PropTypes.bool,
+  active: PropTypes.bool,
+  disabled: PropTypes.bool,
+  label: PropTypes.string.isRequired
+};
+
+const PaginationLinkUnstyled = createComponent(styles, PaginationLink);
+
+export default applyTheme(PaginationLinkUnstyled, PaginationLinkTheme);
```
tajo: move to index.js

packages/cf-component-pagination/src/PaginationRoot.js
```diff
@@ -0,0 +1,17 @@
...
+    display: theme.display,
+    clear: theme.clear
+  }
+});
+
+const PaginationRootUnstyled = createComponent(paginationRootStyles, 'div');
+
+export { PaginationRootUnstyled };
+
+export default applyTheme(PaginationRootUnstyled, PaginationRootTheme);
```
tajo: move to index.js

packages/cf-component-pagination/src/PaginationLink.js
sejoker: I use `export default applyTheme(PaginationLinkUnstyled, PaginationLinkTheme);` to import into PaginationItem, I don't want duplication in index.js and PaginationLink

packages/cf-component-pagination/src/PaginationRoot.js
sejoker: I use `export default applyTheme(PaginationRootUnstyled, PaginationRootTheme)` to import into Pagination, I don't want duplication in index.js and PaginationRoot

packages/cf-component-pagination/src/Pagination.js
sejoker: I want consistency with my code, I use default export for PaginationLink and PaginationRoot

packages/cf-component-pagination/src/PaginationItem.js
sejoker: I want consistency with my code, I use default export for PaginationLink and PaginationRoot

packages/cf-component-pagination/src/PaginationLink.js
tajo: you could import them from `./index.js` but that's ok

thejameskyle:
I agree with @tajo that the export stuff should be done in `index.js`, if you're super consistent about modules it will make working with the AST much much easier for the styleguide and any transformations/codemods you want to do.

packages/cf-component-pagination/example/basic/component.js
sejoker: fixed

sejoker:
All right, I moved all exports into `index.js` as per suggestions from @tajo  and @thejameskyle. Should be good to go?

packages/cf-builder-pagination/test/PaginationBuilder.js
```diff
@@ -3,101 +3,116 @@ import renderer from 'react-test-renderer';
...
+    )
   );
   expect(called).toBeTruthy();
   expect(component.toJSON()).toMatchSnapshot();
 });
 
-test('should should call onPageChange when clicking another page', () => {
-  const onPageChange = createStub();
+// StyleProvider does not pass props
+// test('should call onPageChange when clicking another page', () => {
```
thejameskyle: If you're disabling this test you should use `test.skip(` instead so that it appears in test results as being skipped. That way you're reminded to revisit it later.

sejoker: @thejameskyle  valid point, updated

Issue #136, opened 2017-03-29
Remove hardcoded cross dependencies


tajo:
We should change APIs of some of our components. For example, [ModalTitle.js](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-component-modal/src/ModalTitle.js#L5) looks like this:

```jsx
import React, { PropTypes } from 'react';
import { Heading as HeadingUnstyled, HeadingTheme } from 'cf-component-heading';
import { applyTheme } from 'cf-style-container';

const Heading = applyTheme(HeadingUnstyled, HeadingTheme);

class ModalTitle extends React.Component {
  render() {
    return (
      <div className="cf-modal__title">
        <Heading size={3}>
          {this.props.children}
        </Heading>
      </div>
    );
  }
}

ModalTitle.propTypes = {
  children: PropTypes.node
};

export default ModalTitle;
```

The problem is that the styles are hardcoded here and not changeable by consumer. We should change API of ModalTitle.js to

```jsx
import { Heading as HeadingUnstyled, HeadingTheme } from 'cf-component-heading';
import { applyTheme } from 'cf-style-container';
const Heading = applyTheme(HeadingUnstyled, HeadingTheme);

<ModalTitle heading={Heading}>
```

So you can style it outside. Of course, in our case it would be like this:

```jsx
import { Heading } from '@cloudflare/cf-component-heading';
import { ModalTitle } from '@cloudflare/cf-component-modal';

<ModalTitle heading={Heading}>
```

tajo:
Let's tie the themes and components together as default exports in cf-ui. We will deprecate "applyTheme wrappers" in cf-ux.

So for example, the component `cf-component-button` will newly export:

```
Button
ButtonTheme
ButtonUnstyled
ButtonGroup
ButtonGroupTheme
ButtonGroupUnstyled
```

If you use

```js
import { Button } from 'cf-component-button';
```

you will get the styled Button right away. It solves the problem above and makes us more productive. We will still preserve unstyled exports since it costs us nothing and give you a chance to opt out of our theme if you wish.

wyuenho:
Sounds good!

sejoker:
I can't close this issue, because we still need to make change in ModalTitle to pass Heading via props.

PR #138, opened 2017-03-30
Export styled components by default

tajo:
The reasoning is here: https://github.com/cloudflare/cf-ui/issues/136

**TLDR;** Let's make our life easier. No need to ever change APIs because of Fela migration.

Major bumps for: 
- cf-component-button
- cf-component-heading
- cf-component-text

Minor bump for:
- cf-component-modal


hturan:
Awesome! I was worried about the import API incompatibility for the dashboard, thanks for doing this! 💖

Issue #139, opened 2017-03-30
cf-util-http doesn't convert application/json responses to JSON


hturan:
In https://github.com/cloudflare/cf-ui/blob/master/packages/cf-util-http/src/http.js#L64:

```js
body: headers['content-type'] === 'application/json'
      ? JSON.parse(text)
      : text,
```

For responses with a `Content-Type: application/json; charset=UTF-8`, this will fail. We should check that the `Content-Type` header *contains* `'application/json'`.

koddsson:
This got fixed in #141

PR #140, opened 2017-03-30
Fix JSON content-type detection in 'wrapResponse'

hturan:
This fixes an issue described in https://github.com/cloudflare/cf-ui/issues/139 where our detection of whether to parse a response `body` as JSON relied on strict string matching, rather than emulating how it's handled in `superagent`.

The practice for bumping the version isn't defined in CONTRIBUTING, so I won't include it here, but 🙏 *please please please please* 🙏 publish as a `PATCH`, as per the [Semver FAQ](http://semver.org/#what-if-i-inadvertently-alter-the-public-api-in-a-way-that-is-not-compliant-with-the-version-number-change-ie-the-code-incorrectly-introduces-a-major-breaking-change-in-a-patch-release).

### DIFF
--- a/packages/cf-util-http/src/http.js
+++ b/packages/cf-util-http/src/http.js
@@ -29,6 +29,10 @@ export function clearBeforeSend() {
   beforeSendCallbacks = [];
 }
 
+function isJSON(contentType) {
+  return /[\/+]json\b/.test(contentType);
+}
+
 function toQueryParams(kvs) {
   const queryParams = [];
   // Clones the input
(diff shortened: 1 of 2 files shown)


hturan:
Note that this doesn't ensure that `cf-util-http` fully conforms to the `superagent` API, there's a [lot more work to be done there](https://github.com/visionmedia/superagent/blob/21ce90ec63f33827e7208131dd36a9ac8c6c7d00/lib/client.js). This just allows people shipping product internally at Cloudflare to keep working.

koddsson:
>rather than emulating how it's handled in superagent.

Is this patch copy-pasting from `superagent`?

hturan:
> Is this 

[Yeah](https://github.com/visionmedia/superagent/blob/21ce90ec63f33827e7208131dd36a9ac8c6c7d00/lib/client.js#L245), it's using their regex. I've tested it with a few potential `Content-Types` and it seems to work fine.

PR #142, opened 2017-03-31
Forwardport version bumps from legacy

wyuenho:
I needed the modal title styles to be fixed, so I backported master to legacy and made a release

```
 * [new tag]         cf-builder-table@2.2.1 -> cf-builder-table@2.2.1
 * [new tag]         cf-component-button@5.0.0 -> cf-component-button@5.0.0
 * [new tag]         cf-component-heading@4.0.0 -> cf-component-heading@4.0.0
 * [new tag]         cf-component-modal@6.1.0 -> cf-component-modal@6.1.0
 * [new tag]         cf-component-text@4.0.0 -> cf-component-text@4.0.0
```

In a previous release, Kristjan also minor bumped cf-component-card, so I'm making this PR to keep legacy and master in sync one last time. In the future, new feature not related to fela or bug fixes should be done in the legacy branch. If you make a change in "legacy", it your responsibility to make a separate PR to forwardport that change back to master. Also, version bumps in the legacy branch should only involve minor and patch bumps, although exception may be made on a case-by-case basis. The master branch is reserved for fela migration. Releases from there should only involve major bumps.


tajo:
@wyuenho Can you merge?

sejoker:
@wyuenho we should consider using git rebase over squash strategy as a default. The main drawback of using squash - we loose the original commits history. Git blame becomes inefficient. As an example: I didn't know my changes in `cf-util-http` is already merged into master. It wasn't obvious because of git squash.

Thoughts? cc @tajo @jwineman.

jwineman:
I avoid squash cause I like verbose commit histories.  I know some  people hate it though, I don't feel strongly.

PR #144, opened 2017-04-01
Styling component similar to cf-component-flex

jculvey:
We're embarking on a massive overhaul of some recent feature work and I'd like to clean house while we have the chance. I'd like to avoid the further proliferation of div and span with custom css in our growing codebase. 

This is sort of a stop gap as we implement fela in all of our cf-ui components and work out the kinks in usage, which may take a little time. The cf-component-flex overhaul was super awesome, and fixed most of the styling problems we hit in our recent LB work.

My thinking is that replacing the usage of a single component in several places will be easier to undo than transitioning a ton of css to css-in-js. To make this component conducive to that, the interface is identical to a fela rule.

I'm not super fond of the name box, but I could really thing of a better name. <antidiv/> perhaps? 

As an added benefit, we can add some safety for developers by checking for styling gotchas, should they attempt to do reckless things.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,8 @@
     "react-test-renderer": "^15.4.2",
     "redux": "^3.6.0",
     "redux-thunk": "^2.2.0",
-    "rimraf": "^2.6.1"
-  }
+    "rimraf": "^2.6.1",
+    "babel-plugin-polished": "^1.0.3"
+  },
+  "dependencies": {}
 }
--- a/packages/cf-component-box/.npmignore
+++ b/packages/cf-component-box/.npmignore
@@ -1,0 +1,6 @@
+node_modules
+*.log
+src
+coverage
+test
+example
--- a/packages/cf-component-box/README.md
+++ b/packages/cf-component-box/README.md
@@ -1,0 +1,37 @@
+# cf-component-box
+
+> Cloudflare Box Component
+
+This is styling component similar to cf-component-flex which allows for arbitrary styling.
+
+This component should only be used when the built in style doesn't quite fit the bill and when
+it can't be done with fela. Ideally, cf-ui components should have the right styling built in for
+most cases. In the few edge cases that pop up in real world feature development, this is an alternative
+to wrapping your component in a `<div>` or `<span>` and styling it with css to accomodate one-offs.
+
+Also, note that using Box effectively performs a component level CSS reset, as all props
+not provided will be pulled from the default theme. These defaults, outlined in propertiesSpec,
+either match the browser defaults, or are taken from our global styles as defined by
+cf-style-const.
+
+## Installation
+
+```sh
+$ npm install cf-component-box
+```
+
+## Usage
+
+```jsx
+import React from 'react';
+import { Box } from 'cf-component-box';
+
+const SomeComponent = () => (
+  <Box margin="1rem" padding="1rem">
+    <Heading size={3}>Things</Heading>
+    <Text>Lorem ipsum</Text>
+  </Box>
+);
+
+export default SomeComponent;
+```
--- a/packages/cf-component-box/example/basic/component.js
+++ b/packages/cf-component-box/example/basic/component.js
@@ -1,0 +1,100 @@
+import React from 'react';
+import { Box } from 'cf-component-box';
+
+const BoxComponent = () => (
+  <Box>
+    <Box margin="1rem" backgroundColor="thistle">
+      This is a box with a margin
+    </Box>
+
+    <Box padding="1rem" backgroundColor="blanchedalmond">
+      This is a box with padding
+    </Box>
+
+    <Box
+      padding="1rem"
+      marginTop="1rem"
+      borderTopColor="black"
+      borderTopWidth={1}
+      borderTopStyle="solid"
+      borderBottomColor="blue"
+      borderBottomWidth={1}
+      borderBottomStyle="groove"
+      borderLeftColor="teal"
+      borderLeftWidth={2}
+      borderLeftStyle="dashed"
+      borderRightColor="purple"
+      borderRightWidth={2}
+      borderRightStyle="dotted"
+    >
+      All the borders!
+    </Box>
+
+    <Box
+      display="inline-block"
+      margin="1rem"
+      padding="1rem"
+      width="40%"
+      textAlign="center"
+      backgroundColor="#3dd"
+    >
+      inline block
+    </Box>
+
+    <Box
+      margin="1rem"
+      padding="2rem"
+      backgroundColor="peachpuff"
+      display="inline-flex"
+      justifyContent="space-between"
+      width="40%"
+    >
+      <Box backgroundColor="lightsalmon" margin="1rem" padding="1rem">Flex</Box>
+      <Box backgroundColor="lightsalmon" margin="1rem" padding="1rem">Flex</Box>
+      <Box backgroundColor="lightsalmon" margin="1rem" padding="1rem">Flex</Box>
+    </Box>
+
+    <Box marginLeft="1rem">
+      <h4>Absolute Positioning</h4>
+    </Box>
+
+    <Box
+      marginLeft="1rem"
+      backgroundColor="aquamarine"
+      position="relative"
+      width={200}
+      height={200}
+      display="inline-block"
+    >
+      <Box
+        backgroundColor="pink"
+        position="absolute"
+        width={50}
+        height={50}
+        top={10}
+        left={10}
+      />
+      <Box
+        margin="1rem"
+        backgroundColor="azure"
+        position="absolute"
+        width={100}
+        height={100}
+        bottom={10}
+        right={10}
+      />
+    </Box>
+
+    <Box
+      margin="1rem"
+      padding="1rem"
+      backgroundColor="lightsalmon"
+      float="right"
+    >
+      Floated right
+    </Box>
+
+  </Box>
+);
+
+export default BoxComponent;
--- a/packages/cf-component-box/package.json
+++ b/packages/cf-component-box/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "cf-component-box",
+  "description": "Cloudflare Box Component",
+  "version": "1.0.0",
+  "main": "lib/index.js",
+  "module": "es/index.js",
+  "jsnext:main": "es/index.js",
+  "author": "James Culveyhouse <jculveyhouse@cloudflare.com>",
+  "license": "BSD-3-Clause",
+  "publishConfig": {
+    "registry": "http://registry.npmjs.org/"
+  },
+  "dependencies": {
+    "cf-style-const": "^0.3.1",
+    "cf-style-container": "^0.2.1",
+    "polished": "^1.0.2"
+  },
+  "peerDependencies": {
+    "react": "^0.14.0 || ^15.0.0-0 || ^16.0.0-0"
+  }
+}
--- a/packages/cf-component-box/src/Box.js
+++ b/packages/cf-component-box/src/Box.js
@@ -1,0 +1,141 @@
+import React, { PropTypes } from 'react';
+import { createComponent } from 'cf-style-container';
+import propertiesSpec from './propertiesSpec';
+
+const propertyNames = Object.keys(propertiesSpec);
+
+// Generates styles given the provided props.
+//
+// This is driven by the metadata provided in propertiesSpec
+// and produces a data structure like this:
+//
+// {
+//   margin: props.margin || props.theme['margin']
+//   width: props.width || props.theme['width']
+//   ...
+// }
+//
+// See propertySpec.js for more detail
+//
+const styles = props => {
+  let styles = {};
+
+  const propKeys = Object.keys(props);
+
+  propKeys.forEach(name => {
+    // Lookup the spec for the property
+    const spec = propertiesSpec[name];
+
+    // Ignore non-style props
+    if (spec === undefined) {
+      return;
+    }
+
+    // Get the style value from props
+    let value = props[name];
+
+    // pre-process the value with all pre-processors
+    if (spec.preprocessWith) {
+      spec.preprocessWith.forEach(fn => {
+        value = fn(value);
+      });
+    }
+
+    styles[name] = value;
+  });
+
+  // Only do checks in dev mode
+  if (process.env.NODE_ENV !== 'production') {
+    checkForViolations(styles);
+  }
+
+  return styles;
+};
+
+const Box = props => {
+  const { className, children } = props;
+
+  // Render the div with the fela classname
+  return (
+    <div className={className}>
+      {children}
+    </div>
+  );
+};
+
+const checkForViolations = props => {
+  // Check for conflicts given the spec's conflictsWith property
+  // See propertySpec.js for more detail
+  propertyNames.forEach(propName => {
+    const spec = propertiesSpec[propName];
+    if (!props[propName] || !spec.conflictsWith) return;
+
+    spec.conflictsWith.forEach(otherPropName => {
+      if (props[propName] && props[otherPropName]) {
+        const msg = `${propName} cannot be set in conjunction with ${otherPropName}`;
+        console.error(msg);
+      }
+    });
+  });
+
+  // TODO: To pull this check off, we'll need re-evaluate our baseline.
+  // current is 22.5px
+  // checkLineHeightViolations(props);
+};
+
+const checkLineHeightViolations = props => {
+  const rem2px = rem => {
+    return parseFloat(rem) * props.fontSize;
+  };
+
+  const baseline = props.lineHeight * props.fontSize;
+
+  // Calculate total margin, padding, and border
+  let totalMargin = 0;
+  if (props.margin) {
+    totalMargin = rem2px(props.margin);
+  } else {
+    totalMargin += rem2px(props.marginTop) || 0;
+    totalMargin += rem2px(props.marginBottom) || 0;
+  }
+
+  let totalPadding = 0;
+  if (props.padding) {
+    totalPadding = rem2px(props.padding);
+  } else {
+    totalPadding += rem2px(props.paddingTop) || 0;
+    totalPadding += rem2px(props.paddingBottom) || 0;
+  }
+
+  let totalBorder = 0;
+  if (props.borderHeight) {
+    totalBorder = rem2px(props.borderHeight);
+  } else {
+    totalBorder += rem2px(props.borderTopHeight) || 0;
+    totalBorder += rem2px(props.borderBottomHeight) || 0;
+  }
+
+  // Warn the user if the total spacing isn't a multiple of the baseline
+  const spacing = totalMargin + totalBorder + totalPadding;
+  if (spacing && baseline % spacing !== 0) {
+    console.warn(
+      `Spacing from padding, border, and margin surrounding element should be a multiple of baseline ${baseline}.`
+    );
+  }
+};
+
+// Extract proptypes from the propertiesSpec
+const extractPropTypes = () => {
+  let propTypes = {};
+
+  // Loop through all property specs and pluck the prop type
+  propertyNames.forEach(name => {
+    propTypes[name] = propertiesSpec[name].propType;
+  });
+
+  return propTypes;
+};
+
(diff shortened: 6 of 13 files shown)


packages/cf-component-box/example/basic/component.js
```diff
@@ -0,0 +1,71 @@
...
+
+    <Box
+      padding="1rem"
+      marginTop="1rem"
+      borderTopColor="black"
+      borderTopWidth="1"
+      borderTopStyle="solid"
+
+      borderBottomColor="blue"
+      borderBottomWidth="1"
```
tajo: numbers should be numbers`borderBottomWidth={1}` ?

packages/cf-component-box/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "cf-component-box",
+  "description": "Cloudflare Box Component",
+  "version": "3.4.1",
```
tajo: What happened to the previous versions? :)

packages/cf-component-box/src/Box.js
```diff
@@ -0,0 +1,117 @@
...
+import { createComponent } from 'cf-style-container';
+
+const styles = (props) => ({
+  margin: props.margin || props.theme['margin'],
+  marginTop: props.marginTop || props.theme['marginTop'],
+  marginBottom: props.marginBottom || props.theme['marginBottom'],
+  marginLeft: props.marginLeft || props.theme['marginLeft'],
+  marginRight: props.marginRight || props.theme['marginRight'],
+
+  padding: props.padding || props.theme['padding'],
```
tajo: This will lead to bugs if you define both `padding` and `paddingTop` at the same time. Fela can't enforce the order of properties because of [atomic CSS design](http://fela.js.org/docs/introduction/Drawbacks.html). We can't mix longhand and shorthand properties. It needs to be one of either.

packages/cf-component-box/example/basic/component.js
```diff
@@ -0,0 +1,71 @@
...
+    <Box display="inline-block" margin="1rem" width="400" backgroundColor="#3dd">
+      half box
+    </Box>
+
+    <Box display="inline-block" margin="1rem" width="400" backgroundColor="#3dd">
+      half box
+    </Box>
+
+    <Box
+      margin="1rem"
```
tajo: Idea: Should we automatically convert pixels to rems here if used. I use this handy helper https://polished.js.org/docs/#rem. Is there a valid use for pixels nowadays?

```js
<Box margin={15} />
```

would be converted to

```js
<div margin="1rem">
```

in our case where fontSize is 15.

packages/cf-component-box/src/Box.js
```diff
@@ -0,0 +1,117 @@
...
+  top: props.top || props.theme['top'],
+  bottom: props.bottom || props.theme['bottom'],
+  left: props.left || props.theme['left'],
+  right: props.right || props.theme['right'],
+
+  borderWidth: props.borderWidth || props.theme['borderWidth'],
+  borderColor: props.borderColor || props.theme['borderColor'],
+  borderStyle: props.borderStyle || props.theme['borderStyle'],
+
+  borderTopWidth: props.borderTopWidth || props.theme['borderTopWidth'],
```
tajo: I would like to see some vertical rhythm checker even in this first version (summing up border widths with margins and paddings) and maybe giving some `console.warn` if it doesn't equal to "base line-height" * X? Just to give an idea to other devs how powerful this `<Box/>` can be in the future for our design, so they get a strong motivation to use it.

packages/cf-component-box/example/basic/component.js
jculvey: Yeah totally. I tried to used polished but its a gigantic thing to pull in for 1 function :/

packages/cf-component-box/package.json
jculvey: They're hanging out with ipv5

packages/cf-component-box/src/Box.js
jculvey: Good point. I added checking for this via the 'conflictsWith' property in the propertySpec. It should throw an error now if you do this.

packages/cf-component-box/src/Box.js
jculvey: I wrote some placeholder code, but this didn't quite pan out because of our numbers :/

Our current baseline is 22.5 pixels, which really sucks. It's also encoded as a sass variable. I think there's some benefit to encoding this in a single source of truth, perhaps in yet another package that exports a simple data structure.

packages/cf-component-box/example/basic/component.js
jculvey: Yes :)

jculvey:
@tajo Great points. I basically re-wrote what I had to make it easier to support all the things.

I added all the ones I could find from that file you pointed me to the other day. There's now support for all flexbox props, zIndex, textAlign, positioning, and a bunch of other stuff.

I'm sure i've missed some, but its super easy to add new ones now by adding new entries in propertiesSpec.

packages/cf-component-box/src/propertiesSpec.js
```diff
@@ -0,0 +1,342 @@
+import React, { PropTypes } from 'react';
+
+const numberOrString = PropTypes.oneOfType([
+  PropTypes.string,
+  PropTypes.number
+]);
+
+// TODO: This needs to be pulled from a proper single source of truth
+const baseFontSize = 15;
+const pxToRem = pixels => {
```
tajo: I would rather use the helper from polished so we can easily compile it away with https://github.com/styled-components/babel-plugin-polished

packages/cf-component-box/src/Box.js
```diff
@@ -0,0 +1,141 @@
...
+
+  // Render the div with the fela classname
+  return (
+    <div className={className}>
+      {children}
+    </div>
+  );
+};
+
+const checkForViolations = props => {
```
tajo: This should be stripped in production, right?

packages/cf-component-box/src/Box.js
```diff
@@ -0,0 +1,141 @@
...
+//   ...
+// }
+//
+// See propertySpec.js for more detail
+//
+const styles = props => {
+  let styles = {};
+
+  const propertyNames = Object.keys(propertiesSpec);
+  propertyNames.forEach(name => {
```
tajo: Do we really need a for loop? I'm a bit worried about performance. We should probably avoid stuff like this considering how often this component will be used.

packages/cf-component-box/src/propertiesSpec.js
```diff
@@ -0,0 +1,342 @@
+import React, { PropTypes } from 'react';
+
+const numberOrString = PropTypes.oneOfType([
+  PropTypes.string,
+  PropTypes.number
+]);
+
+// TODO: This needs to be pulled from a proper single source of truth
+const baseFontSize = 15;
```
tajo: https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-const/src/variables.js#L7

packages/cf-component-box/src/index.js
```diff
@@ -0,0 +1,7 @@
+import { applyTheme } from 'cf-style-container';
+import BoxUnstyled from './Box';
+import BoxTheme from './BoxTheme';
+
+const Box = applyTheme(BoxUnstyled, BoxTheme);
+
+export { BoxUnstyled, Box };
```
tajo: export BothTheme as well

packages/cf-component-box/src/Box.js
jculvey: Yeah, whats the mechanism I should use to do this?

packages/cf-component-box/src/propertiesSpec.js
jculvey: oh yay!

packages/cf-component-box/src/propertiesSpec.js
jculvey: Cool, I'll add this.

packages/cf-component-box/src/Box.js
jculvey: I don't like the for loop either. This is the only place that bothers me, since this is the only one that will get hit in a production build. With the current functionality, I can't think of a good way to avoid it. 

If we're ok with dropping the behavior of providing a default if a prop isn't passed to the component, the loop can be much better by looping over the prop names and comparing those to all the properties. In most cases the number of props will be small (<10) and we can exit early with no work by checking membership in propertiesSpec.

The other option is to unroll it and write a line per specced property, which is annoying from a dev perspective.

packages/cf-component-box/src/Box.js
thejameskyle:
```js
if (process.env.NODE_ENV !== 'production') {
  ...
}
```

Dead code elimination will handle it

packages/cf-component-box/src/Box.js
```diff
@@ -0,0 +1,139 @@
...
+  } else {
+    totalBorder += rem2px(props.borderTopHeight) || 0;
+    totalBorder += rem2px(props.borderBottomHeight) || 0;
+  }
+
+  // Warn the user if the total spacing isn't a multiple of the baseline
+  const spacing = totalMargin + totalBorder + totalPadding;
+  if (spacing && baseline % spacing !== 0) {
+    console.warn(
+      `Spacing from padding, border, and margin surrounding element should be a multiple of baseline ${baseline}.`
```
thejameskyle: Why not use abstract units that you turn into a multiple of the baseline?

tajo: You mean something like "1x", "2x", "3x" ... ? Sometimes you need arbitrary ratio between top and bottom paddings/margins/borders.

packages/cf-component-box/src/propertiesSpec.js
```diff
@@ -0,0 +1,347 @@
...
+]);
+
+const pxToRem = value => {
+  if (typeof value === 'number' || /\d+px/.test(value)) {
+    return rem(value, variables.fontSize);
+  }
+  return value;
+};
+
+// The export of this module is a datastructuer containing metadata
```
sejoker: data structure

sejoker:
As an alternative name to box I can suggest `block`.  We used to call similar element as a block in my previous project.

packages/cf-component-box/README.md
```diff
@@ -0,0 +1,53 @@
...
+to wrapping your component in a `<div>` or `<span>` and styling it with css to accomodate one-offs.
+
+It supports the following styling properties:
+  * margin
+  * padding
+  * border
+    * borderWidth
+    * borderColor
+    * borderStyle
+  * backgroundColor
```
sejoker: The common use case for the Box could be to display an image. In that case we could consider adding:
- backgroundImage
- backgroundRepeat
- backgroundPositionX
- backgroundImagePositionY

It's definitely could be added as a separate PR, just share my thoughts.

packages/cf-component-box/README.md
```diff
@@ -0,0 +1,53 @@
+# cf-component-box
+
+> Cloudflare Box Component
+
+This is styling component similar to cf-component-flex which allows for arbitrary styling.
+
+This component should only be used when the built in style doesn't quite fit the bill and when
+it can't be done with fela. Ideally, cf-ui components should have the right styling built in for
+most cases. In the few edge cases that pop up in real world feature development, this is an alternative
+to wrapping your component in a `<div>` or `<span>` and styling it with css to accomodate one-offs.
```
koddsson: I'm a bit confused probably from not working with `fela` at all before. Is it not possible just to inline style a div if you need to do customization?

hturan:
This is super interesting! Reminds me of [react-primitives](https://github.com/lelandrichardson/react-primitives).

I guess my main question would be why what does this give us that inline styles or `fela` doesn't? The `cf-component-flex` overhaul is just a nice way to apply inline styles to a `div`. Is this just a similar approach but with *all* styles and using `fela`?

> I'm not super fond of the name box, but I could really thing of a better name. perhaps?

Naming is probably the least important thing here, but for what it's worth it seems most abstract container elements are called `View` (or sometimes `Box`).

packages/cf-component-box/README.md
jculvey: Absolutely. Ideally this should have all possible properties including these.

packages/cf-component-box/README.md
jculvey: Yeah its possible to use div, and thats actually what we're doing now. 

The idea with this is that the fela migration isn't 100% complete yet, and we've all got a bunch of stuff on our plates to either write or refactor. Instead of doing it with div's and either css or inline styles like we've been doing, we'd be using a component that implements all its styling behavior in fela.  It lets us:
   - start using fela today without the migration being complete
   - which lets us use inline styles today with all the perks of fela
   - If we use a single component for these edge cases we can track and replace them easier when the migration is complete.

jculvey:
@hturan good points.

>  Is this just a similar approach but with all styles and using fela?

Yup. That's really all it is.

> A concern would be that we'd miss a layout option (I don't see any mention of float or overflow, for instance) and need to do a whole PR/merge/publish/update cycle.

Yeah I'm sure I've missed some. I'll add the ones that Yevgen and you have mentioned. On the plus side, to support a new property the minimum you must do is add an entry into propertiesSpec with its propType. I don't mind doing this every time if someone pings me.

jwineman:
If we only intend this to be used while we migrate to Fela and we want to rip it out after lets give it a scarier name.  `<Box/>` seems like an _actual_ component.  What about `<DangerouslyInline/>` or `<DangerouslyStyle/>`?

jculvey:
@jwineman it may end up retaining its use after a full migration. @tajo and I discussed using this permanently as a beefier alternative to Flex and FlexInput. Those are both great, but we could use something that supports all display props as well as position props.

jwineman:
after the migration can it be merged with `<Flex>`?  Does it make sense to keep it separate _after_ Fela is complete?

jculvey:
@jwineman Yeah it can be merged, with some work. We'd have to think about the name and interface. Flex has 2 elements, a parent and a child, and Box has only the one element. 

Flex is probably too narrowly named to support all display and position props. I wouldn't want to say <Flex display="inline-block"> or <Flex position="absolute" left={10}>.

jwineman:
If there is a use case for it I'm fine keeping it post migration, I just want to make sure we aren't introducing new components unless we absolutely need to.  

I'm a little confused because I thought the positioning problem was solved by fela - after we finish the migration why can't these styles be applied to the fela themes themselves?

jculvey:
@jwineman Good question actually. @tajo Let's say I have a new component like so:

```jsx
  <SomeComponent>
    <Heading>Foo</Heading>
    <Text>Bar</Text>
  </SomeComponent>
```
If I want to set `display: inline-block` on the Heading and Text components, do they explicitly need to return 'display: theme.display' in their styles function?

If not, how would I go about overriding their display property?

packages/cf-component-box/README.md
tajo: @koddsson please don't use inline styles anymore and try out fela instead, it's exactly like inline styles but you can use everything that you know from CSS (media queries, ::before ...) and it's actually using `<style></style>` so the performance is much better

tajo: You can also do this:

```jsx
import { createComponent } from 'cf-style-container';
const MyDiv = createComponent(props => { color: yellow }, 'div');
```

tajo:
@jculvey The only way how you can (should) change the style of component is through its props. So I would add a new prop `inline={true}` to Heading and Text if needed.

packages/cf-component-box/src/Box.js
thejameskyle: Just provide any number, units of 1, 2, 3 are multiple by say 10. So if you could work off of that.

jculvey:
Thanks @tajo. So @jwineman, I think given that, this box could have long term use. There could be a potential need for adjusting the display, position, border, margin of literally every component. 

Design can give us some firm commitments to anything INSIDE the component, but as for outside of it (will it be used in a card? a table row? a tooltip on a graph?), this will probably vary quite a bit depending on the feature.

So, we either have to explicitly  implement these props for every cf-ui component as @tajo explained above, or we have 1 thing that supports the subset of properties needed when consuming nearly every component.

tajo:
I'm open to introducing a "backdoor" to our cf-ui components. It would be a prop called `overwriteDefaultRules`. You could pass it a [rule](http://fela.js.org/docs/basics/Rules.html) or an object of rules (in case where we use multiple classNames & rules for one React component). Then, it would be combined with the existing default rule via [combineRules](https://github.com/rofrischmann/fela/blob/master/docs/api/fela/combineRules.md). This supports nesting as well.

This is the proper way how to do it if we really want to have an escape hatch. I am not a big fan of using it but if you really really really need to ship and you will not abuse it, then I will rather see this than not using cf-ui components at all and hacking everything through cf-component-box or even worse using bunch of `<div>`s for everything.

jwineman:
`dangerouslyOverwriteDefaultRules`.  

It should be hard to type and visually obvious when you're reading it that its a hack.

wyuenho:
I haven't read thru all the comments, but my guess is, cf-component-flex turns out to be far less useful then we thought. We want to lay out things, but the flexbox properties have to be merged into the inner component markups to be useful and avoid `<div>`ititus.  I think once we've migrated all the cf-ui components, we should be able override styles from the parent, probably thru some mechanism similar to @tajo suggests. For now, just ignore cf-component-flex and use `<div>`. cf-component-flex is also something I wanted to remove for a while now.

jculvey:
@jwineman @tajo 

Here's where I'm at right now. We really have to start moving on some feature work, which 
is going to involve implementing a lot of new things and refactoring a bunch of old ones. I'm 
working on a ticket right now which i've been done with since last week, but it uses inline styles & divs.

I made some effort to use Fela on Friday, but Vojtech and I hit some roadblocks. For one,
if your component has children that also need special styling, there isn't an easy way to define nested style rules to solve this. Rendering one element per component just isn't an option.

Furthermore, there's a lot of properties that aren't supported in every component right now (like layout, border, padding, line-height, color, vertical-align, height).

I have to move the needle on feature work on a weekly basis, and here's the only options I see:

1. Do whatever is needed to allow 100% of my feature's spec to be implemented with only cf-ui components and Fela.
2. Get a minimal version of this box thing checked in and start using it.
3. Just continue to hack it with divs.

I honestly think this PR is a middleground between enforcing a platform that we realistically won't have in our hands for several months and opening pandoras box by allowing people to willy nilly pass any style to any thing. I'm happy to do whatever, but I'd like to know how you guys think I should proceed. I don't want to pollute the waters of something nice "because deadlines", but we all need a way to hit a weekly cadence of feature work even in the short term.

jwineman:
I'm fine with this for now.  My concerns were around not deprecating it after Fela is complete.  We can cross that bridge when we get to it though.

package.json
```diff
@@ -56,5 +56,9 @@
     "redux": "^3.6.0",
     "redux-thunk": "^2.2.0",
     "rimraf": "^2.6.1"
+  },
+  "dependencies": {
+    "babel-plugin-polished": "^1.0.3",
```
tajo: Should be in devDeps?

package.json
```diff
@@ -56,5 +56,9 @@
...
+    "polished": "^1.0.2"
```
tajo: Should be in the cf-component-box package.json ?

packages/cf-component-box/.npmignore
```diff
@@ -0,0 +1,7 @@
+node_modules
+*.log
+src
+coverage
+test
+example
+karma.conf.js
```
tajo: We don't karma anymore. :)

jculvey:
I've updated this with some minor fixes to comments. 

I'm totally open to closing this, or at least shelving it for a while. I'm hesitant to merge this if everyone feels sort of mediocre on it. If we have a cleaner plan for overrides that everyone is comfortable with, it probably makes sense to just go with that and refactor when its ready.

tajo:
This should go just into master. I'm happy with the merge.

jwineman:
Hes putting this into WWW-next common for now.  If everyone likes it we can merge this PR.

wyuenho:
Why is this merged? I thought we were moving this to www-next for incubation?

PR #145, opened 2017-04-03
Styleguide: Use styleguidist for a styleguide

koddsson:
Add a dependency to styleguidist and generate some of our documentation with it.

styleguidist seems like a pretty good project that generates a beautiful interactive styleguide and I think we should use it instead of our homebuilt one.

I figured I'd open up this PR early with a small example but I'd want to go in and change all the components to have examples as well as loading the correct fela theme in the examples.


koddsson:
I think it's pretty neat!

image

hturan:
[react-styleguideist](https://github.com/styleguidist/react-styleguidist) is super cool and does a lot of the things we're looking to do with our documentation (using react-docgen to parse `propTypes` as a webpack loader, etc).

In my eyes, what this PR gives us is more useful than the documentation that we currently ship, and it doesn't even currently include documentation.

koddsson:
>get it to output static HTML for GitHub Pages

There's a option for that :)

tajo:
This is good stuff! It does pretty much everything I wish to see from cf-styleguide in the short term. However, I think the right approach here is to "steal" parts of it and add it to cf-styleguide than 100% commit to it and rewrite all our examples.

There are already some things in cf-styleguide that you can't replace by 3rd party tool:
- it directly loads source codes from cf-ui and cf-ux (super useful for deving)
- it loads also our old postcss styles
- HMR for React, Fela and legacy postcss
- you can always improve it since you have a complete control

koddsson:
@tajo I feel like I've been missing a lot if this is the case. There's no scripts to run the styleguide in the `cf-ui` package.json and no apparent way to get what's running on cloudflare.github.io/cf-ui locally.

I'd also like to point out that `styleguidist` does have HMR out of the box and I feel like a lot of the postcss arguments are kind of invalid since we're moving to fela anyways.

I think the approach we should be taking is building on OSS since the object of `cf-ui` seems to be components and not a styleguide. 

>you can always improve it since you have a complete control

You can always improve OSS like `styleguidist` and I'd much rather commit upstream to that project than make yet another styleguide that kind of works half of the time.

tajo:
@koddsson

> I feel like I've been missing a lot if this is the case. There's no scripts to run the styleguide in the cf-ui package.json and no apparent way to get what's running on cloudflare.github.io/cf-ui locally.

It now lives in cf-styleguide. It's under the WWW on stash. There is a README describing how to build cloudflare.github.io/cf-ui and how to run it in dev mode.

> I'd also like to point out that styleguidist does have HMR out of the box and I feel like a lot of the postcss arguments are kind of invalid since we're moving to fela anyways.

Will HMR work with Fela? How do you even add an extra `style` node for Fela to make it work? Why does it use markdown to create examples? Will we always have to wrap all examples in a boilerplate code that sets up Fela?

> I think the approach we should be taking is building on OSS since the object of cf-ui seems to be components and not a styleguide.

That's why we removed it from cf-ui and it's now in cf-styleguide. So far it's more a documentation/examples than real "style guide" but the idea is that we'll give it some organization and proper design so there will be sections full of texts talking about design stuff. It will have 3 functions:

- design principles (style guide)
- cf-ui, cf-ux documentation
- editor for developers (HMR for both styles and code)

We can keep everything in sync and one place.

> You can always improve OSS like styleguidist and I'd much rather commit upstream to that project than make yet another styleguide that kind of works half of the time.

That's true. I like OSS a lot. The problem with projects like styleguidist is that it needs to work for everyone, so it must do a lot of hard assumptions and it will be impossible in many cases to bring up our local use-cases. We already went this path with our documentation. 

Do you really want to go [this path](https://github.com/styleguidist/react-styleguidist/blob/master/examples/customised/styleguide.config.js
)?

Also, it's not like `styleguidist` gives you everything for free. You are about to write 100 code examples in markdown with a lot of boilerplate. Maybe there is a better way. If we are in control of both ends (components and style guide) can we auto-generate things?

I'm happy to setup a meeting and chat about this. It seems there is a big gap in communication. It seems that nobody really understands how Fela works except me and Yevgen, why do we even want to use it and what work already came into the tooling and architecture.

koddsson:
Oh ok. Can we put this new styleguide on [cloudflare.github.io/cf-ui](cloudflare.github.io/cf-ui)? I'd love to play with the editor you mentioned, the one that comes with `styleguidist` is one of my favorite features. Also could you throw me a Button fela example? [Non of the documentation ones seems to work for me.](https://esnextb.in/?gist=6b251c438598f9cbd34e786c9c12a917)

sejoker:
I tend to agree with @tajo we should integrate `good parts` from styleguidist project into our cf-styleguide instead of switching to the new shiny thing without evaluating all the benefits and shortcomings first. All in all I appreciate the idea of making our styleguide `great again`.

koddsson:
Thanks! Maybe it's worth updating the examples with this? The `cssbeautify` doesn't seem to be browserify specific since @hturan gets the same [with webpack](https://www.webpackbin.com/bins/-Kgsy96y7R1AQOdet_5U)

koddsson:
I just think it's gonna be more working creating something on our own when we can adopt the tools already out there.

package.json
```diff
@@ -22,9 +22,12 @@
     "prettier": "node prettier.js",
     "clean": "lerna clean --yes && rimraf .jest",
     "publish": "git pull --rebase && check-installed-dependencies && npm run build && lerna publish",
-    "update-snapshot": "cross-env BABEL_ENV=commonjs jest --no-cache --updateSnapshot"
+    "update-snapshot": "cross-env BABEL_ENV=commonjs jest --no-cache --updateSnapshot",
+    "styleguide": "styleguidist server"
   },
   "devDependencies": {
+    "@webpack-blocks/babel6": "^0.4.1",
+    "@webpack-blocks/webpack2": "^0.4.0",
```
wyuenho: What is this scope?

koddsson: https://github.com/andywer/webpack-blocks

koddsson: Found it in the docs for react-styleguidist https://github.com/styleguidist/react-styleguidist/blob/master/docs/Webpack.md#non-webpack-projects

koddsson:
@tajo A lot of comments. I'll try to get through them all 🙂

>> I'd also like to point out that styleguidist does have HMR out of the box and I feel like a lot of the >postcss arguments are kind of invalid since we're moving to fela anyways.

>Will HMR work with Fela?
>How do you even add an extra `style` node for Fela to make it work?
>Why does it use markdown to create examples?
>Will we always have to wrap all examples in a boilerplate code that sets up Fela?

- I've added the styles for the button in 13ab35c.
- Are you talking about the `<style />` tag in the `<head />` so Fela can inject styles? I overwrote the handlebar template in 13ab35c to include it.
- Markdown adds more depth to the examples since you can add descriptions, multiple code examples and general markup. All the code blocks will be interpreted as react components.
- Yeah. We can write some code to abstract this away if we'd like, but I personally prefer to have examples copy-pasteable.

>> I think the approach we should be taking is building on OSS since the object of cf-ui seems to be components and not a styleguide.

>That's why we removed it from cf-ui and it's now in cf-styleguide. So far it's more a >documentation/examples than real "style guide" but the idea is that we'll give it some organization >and proper design so there will be sections full of texts talking about design stuff. It will have 3 >functions:

>- design principles (style guide)
>- cf-ui, cf-ux documentation
>- editor for developers (HMR for both styles and code)

>We can keep everything in sync and one place.

Maybe we have a bit of a disconnect here in what we determine a styleguide. Maybe we should think about it as two things. One being a design guide and the other documentation about our components. I'm less interested in the design guide aspect and much more interested in documentation for our components so I can use them and understand how they work when working on the web interface.

`styleguidist` provides pretty good `cf-ui` documentation as well as a editor for developers. Give the editor a spin if you haven't already. It's pretty neat!

>> You can always improve OSS like styleguidist and I'd much rather commit upstream to that >>project than make yet another styleguide that kind of works half of the time.

>That's true. I like OSS a lot. The problem with projects like styleguidist is that it needs to work for >everyone, so it must do a lot of hard assumptions and it will be impossible in many cases to bring >up our local use-cases. We already went this path with our documentation. 

I guess here is where our communication does break down a bit. I'm very unaware of our local use-cases. I got the button example up and running pretty easily but I figure that might be a trivial example 😅

Could you list up some of our local use cases and I can try to test this PR against those?

>Do you really want to go [this path](https://github.com/styleguidist/react-?>styleguidist/blob/master/examples/customised/styleguide.config.js
>)?

I wouldn't mind it one bit 😊 I'm actually not sure what you don't like about it.

>Also, it's not like `styleguidist` gives you everything for free. You are about to write 100 code >examples in markdown with a lot of boilerplate. Maybe there is a better way. If we are in control of >both ends (components and style guide) can we auto-generate things?

No matter what way we go, we're always going to have to write up our examples. Using `styleguidist` is the path for less code since we don't have to write thousands of lines of code to get something up and running similar to it.

Not quite sure about the auto-generation. If you're talking about the examples I think we can absolutely do a one-off initial auto generation of the examples.

>I'm happy to setup a meeting and chat about this.

I really like async github threads if you don't mind. It's more transparent conversation and I can better gather my thoughts and put 'em to words. It also produces documentation about our decision making.

>It seems there is a big gap in communication. It seems that nobody really understands how Fela works except me and Yevgen, why do we even want to use it and what work already came into the tooling and architecture.

I agree with the communication. Maybe we can do better by helping each other out to figure out what's needed and what's been done already. A lot of the effort in this PR comes from me wanting to try components before using them in production and figuring out their prop types. `styleguidist` enables me to do that in an almost effortless way and I'm going to continue using it. I figured a lot of you and the people that use our components would like it too 😊

tajo:
> Yeah. We can write some code to abstract this away if we'd like, but I personally prefer to have examples copy-pasteable.

We actually don't want to copy&paste fela boilerplate since it's already done in the root of www-next. We also don't want to use `applyTheme()` but use the default styled exports. So your example works for outsiders but doesn't for CF projects.

> styleguidist provides pretty good cf-ui documentation as well as a editor for developers. Give the editor a spin if you haven't already. It's pretty neat!

Agree, it's really neat.

> Could you list up some of our local use cases and I can try to test this PR against those?

**I have 3 things in my mind right now**
- The `npm run styleguide:build` doesn't server-side render the documentation. The current doc is server-side rendered. I like that it keeps us accountable to write universal components which is important since some of our apps (api.cloudflare.com) are consuming cf-ui and already using server-side rendering. Hopefully even www-next will server-side render one day.
- We want to have this type of documentation for cf-ux (private components) as well. I guess the only option here would be to duplicate styleguidist setup twice which is not ideal.
- We use this concept of themes. I would like to see a browser editor for them as well. It would be especially useful for the global theme. You could change `colorGrayLight` and directly see what and where changes. As a dev I can do it now thanks to HMR but it would be awesome to have it in the browser so you don't have to run the whole stack (designers would like it). You could even add a color picker...

> Using styleguidist is the path for less code since we don't have to write thousands of lines of code to get something up and running similar to it.

Give me a few days. I'll make some improvements to cf-styleguide. I went through the code of styleguidist. The props documentation is using `react-docgen` and this [loader](https://github.com/styleguidist/react-styleguidist/blob/master/loaders/props-loader.js). The editor is powered by [CodeMirror](http://codemirror.net/). My guess is 187 lines of code. :)

koddsson:
@tajo Sounds good! I'm excited to see what comes out of it. I think I might try to address some of your points in `styleguidist` by either contributing upstream to them or maintaining a `react-fela-styleguidist` fork.

I think if I can add themeing and abstract the fela provider out of the examples this could get to a very good spot.

koddsson:
So I'm messing around, mostly to understand fela better. Do you know why I'm not getting any styles with the `ThemeProvider` in this example:

```
const { createRenderer } = require('cf-style-provider');
const { ThemeProvider } = require('cf-style-container');
const { Provider } = require('react-fela');
const { variables } = require('cf-style-const');

const fontNode = document.querySelector('#fela-fonts')
const renderer = createRenderer({ fontNode });

<Provider renderer={renderer} mountNode={fontNode}>
  <ThemeProvider theme={variables}>
    <Button type="primary" onClick={() => console.log('clicked')}>
      Hello world
    </Button>
  </ThemeProvider>
</Provider>
```

I have success with the following example:

```
const { StyleProvider } = require('cf-style-provider');
const { applyTheme } = require('cf-style-container');
const ButtonTheme = require('./src/ButtonTheme').default;
const StyledButton = applyTheme(Button, ButtonTheme);

const cssNode = document.querySelector('#fela-style');
const fontNode = document.querySelector('#fela-fonts');

<StyleProvider
  cssNode={cssNode}
  fontNode={fontNode}
  dev={true}
  selectorPrefix="cf-"
>
  <StyledButton type="primary" onClick={() => console.log('clicked')}>
    Hello world
  </StyledButton>
</StyleProvider>
```

tajo:
Try

```jsx
const { createRenderer } = require('cf-style-provider');
const { ThemeProvider } = require('cf-style-container');
const { Provider } = require('react-fela');
const { variables } = require('cf-style-const');
const { Button } = require('cf-component-button');

const fontNode = document.querySelector('#fela-fonts')
const cssNode = document.querySelector('#fela-css')
const renderer = createRenderer({ fontNode });

<Provider renderer={renderer} mountNode={cssNode}>
  <ThemeProvider theme={variables}>
    <Button type="primary" onClick={() => console.log('clicked')}>
      Hello world
    </Button>
  </ThemeProvider>
</Provider>
```
You need to add this to the header:

```html
<style id="fela-css"></style>
```

Fela is using a different node and renderer for fonts in order to prevent their refetching. The current workflow for fonts is kinda awkward right now. Hopefully it [gets better](https://github.com/rofrischmann/fela/issues/238).

koddsson:
Nice! Thanks for the snippet! Since you're importing Button directly from the `cf-component-button` package it presumably is the "styled" button ie. the theme has been applied to it.

tajo:
Yep. You can find it [here](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-component-button/src/index.js#L8).

PR #146, opened 2017-04-03
Fixes cf-util-http in Edge

sejoker:
avoid use of Headers.entries() and for .. of because of limited support in Edge browser.

more details: https://developer.mozilla.org/en-US/docs/Web/API/Headers\#Browser_compatibility

### DIFF
--- a/packages/cf-util-http/src/http.js
+++ b/packages/cf-util-http/src/http.js
@@ -128,15 +128,16 @@ export function request(method, url, opts, callback) {
   callback = opts.callback;
 
   // Normalize the headers
-  opts.headers = new Headers(opts.headers || {});
+  opts.headers = Object.assign({}, opts.headers);
 
   // Emuluate superagent's questionable ability to filter out headers that's
   // been set to null or undefined.
-  for (let [k, v] of opts.headers.entries()) {
-    if (v === 'null' || v === 'undefined') {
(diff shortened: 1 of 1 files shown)


packages/cf-util-http/src/http.js
```diff
@@ -128,15 +128,16 @@ export function request(method, url, opts, callback) {
...
-  opts.headers = new Headers(opts.headers || {});
+  opts.headers = Object.assign({}, opts.headers);
 
   // Emuluate superagent's questionable ability to filter out headers that's
   // been set to null or undefined.
-  for (let [k, v] of opts.headers.entries()) {
-    if (v === 'null' || v === 'undefined') {
-      opts.headers.delete(k);
+  for (const h in opts.headers) {
+    if (opts.headers[h] === null || opts.headers[h] === undefined) {
```
wyuenho: `opts.headers[h] == null` will get you both cases

packages/cf-util-http/src/http.js
```diff
@@ -162,7 +163,7 @@ export function request(method, url, opts, callback) {
 
   return fetch(url, opts)
     .then(response => {
-      const headers = toHash(response.headers);
+      const headers = toHash(new Map(response.headers));
```
wyuenho: Fucking Microsoft...

PR #147, opened 2017-04-04
Spaced ButtonGroup

tajo:
The original button CSS styles had this rule:

```css
'& + &': {
  marginLeft: '0.4rem'
}
```

If you put two cf-component-buttons next to each other, the right one would get the left margin. However, this selector doesn't work with Fela since all classNames are random and atomic. So we need a different solution.

This PR adds a new prop `spaced` to the component `<ButtonGroup>`, so if you want to have spaces between `<Button>`s wrap them by `<ButtonGroup spaced>...</ButtonGroup>`.

Also, I made the `styles` function a little bit cleaner with `...function()` replacing `Object.assign()`.

screen shot 2017-04-03 at 11 05 07 pm

### DIFF
--- a/packages/cf-component-button/README.md
+++ b/packages/cf-component-button/README.md
@@ -15,17 +15,28 @@ import React from 'react';
 import { Button, ButtonGroup } from 'cf-component-button';
 
 const ButtonComponent = () => (
-  <ButtonGroup>
-    <Button type="primary" onClick={() => console.log('Clicked One!')}>
-      Button One
-    </Button>
-    <Button type="default" onClick={() => console.log('Clicked Two!')}>
-      Button Two
-    </Button>
-    <Button type="success" onClick={() => console.log('Clicked Three!')}>
-      Button Three
-    </Button>
-  </ButtonGroup>
+  <div>
+    <ButtonGroup>
+      <Button type="primary" onClick={() => console.log('Clicked One!')}>
+        Button One
+      </Button>
+      <Button type="default" onClick={() => console.log('Clicked Two!')}>
+        Button Two
+      </Button>
+      <Button type="success" onClick={() => console.log('Clicked Three!')}>
+        Button Three
+      </Button>
+    </ButtonGroup>
+    <p>Button group with spacing</p>
+    <ButtonGroup spaced>
+      <Button type="success" onClick={() => console.log('Clicked Four!')}>
+        Button Four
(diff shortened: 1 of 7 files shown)


sejoker:
as an alternative we could use [monolithic class name](https://github.com/rofrischmann/fela/pull/204/files) and use it for '&+ customclass' styling. It's less code to support and compromise worth to consider.

tajo:
@sejoker yea, this doesn't work with monolithic either since once you add that new property it will change the className of that sibling and they are not considered siblings anymore (different classNames)

Anyway, I agree with Robin that selectors like this should not be used in our component model since they leak styles to other components (siblings in this case). It's more cod but it's cleaner. Also, it's not that common.

You could also do something like

```jsx
<Box><Button /></Box><Box marginLeft="0.4rem"><Button /></Box>
```

in the future, if you don't want to bake it into the cf-ui component.

manatarms:
I just wanted to mention a use case that I encountered. 
I wanted to use the `<Dropdown>` component  which is wrapped by the `<ButtonGroup>` as per our example https://cloudflare.github.io/cf-ui/#cf-component-dropdown. 

I noticed when using only a single button, that on click changes the border-radius on the single button. It should only do this when there is more than one button. I was still forced to use `<ButtonGroup>` because that's the only way to get the dropdown to respect positioning. Just thought this might be relevant and a good time to fix it.

sejoker:
I think it's not a rare use case, in pagination component there is also need for styling sibling element. So it's quite common requirement where you have a list of items.

> this doesn't work with monolithic either since once you add that new property it will change the className of that sibling and they are not considered siblings anymore (different classNames)

I want to play with it, I thought classNames will be static in that case.

wyuenho:
Give me a little bit of time to check how we use Buttons and ButtonGroups. This makes sense to me so far.

sejoker:
I don't have enough time to provide compelling alternative, don't want to block provided solution.

wyuenho:
Yes it's called a button tool bar, although I really don't recall a place where we currently use some variant of this. Bootstrap's button group are spaced between groups, not between buttons.

However,  I believe this was done so we can have contextual margins between buttons, please correct me if I'm wrong. The motivation is slightly different. We are not actually trying to group buttons, we are trying to give spaces between them.

@tajo Please confirm that's the case, if so. This LGTM.

PR #153, opened 2017-04-05
Use yarn for packages bootstrap

sejoker:
- thanks to lerna support of [yarn](https://github.com/lerna/lerna/releases/tag/v2.0.0-beta.38)
- use yarn for bootstrap
- add lock files into packages


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - "6"
 before_script:
-  - yarn run bootstrap
+  - yarn run bootstrap -- --concurrency=1
```
wyuenho: Why is this needed?

wyuenho: Is [this](https://github.com/yarnpkg/yarn/issues/683) still an issue?

sejoker:
I am going to close the PR for the following reasons:
- the locking mechanism is the main benefit of using yarn instead of npm
- we have interdependencies between components
- Lerna should update package's lock file before deploy to npm otherwise old version of the lock file will prevent loading the latest deployed version of interdependency. This feature is not supported at the moment by Lerna

.travis.yml
sejoker: because of [this](https://github.com/lerna/lerna/issues/671)

PR #161, opened 2017-04-06
Back to ES6 classes

wyuenho:
This PR revert functional components back to ES6 classes. The reasons are:

1. We need the refs for form components because we are using redux-form on the product.
2. All of the cf-ui components were in ES6 classes, which gives us refs. Nobody was complaining about them.
3. cf-ui itself is using refs now in [4 different places](https://github.com/cloudflare/cf-ui/search?utf8=%E2%9C%93&q=ref&type=)
4. Refs are needed, when we need to focus or deal with text selection or animation. I'm fine to selectively convert some to functional components if we can identify a list of components that we have a high degree of certainty that we'll never need refs for, which usually are non-interactive components. Form controls, links, tabs, dropdowns, and tables probably need to be in classes. Let's have this discussion first before jumping the gun.
5. We may even be able to use [Pure Components](https://facebook.github.io/react/docs/react-api.html#react.purecomponent) for some of the really simple stuff, assuming we haven't eliminated them by then. Let's ID a list of them first.

We have only migrated 4 components to fela. Two of which are of questionable utility. Let's go back to where it was, have a discussion, then decide what to do.

### DIFF
--- a/packages/cf-component-button/src/Button.js
+++ b/packages/cf-component-button/src/Button.js
@@ -114,25 +114,28 @@ const styles = props => {
   );
 };
 
-const Button = (
-  {
-    onClick,
-    submit,
-    className,
-    disabled,
-    loading,
-    children
+class Button extends React.Component {
+  render() {
+    const {
+      onClick,
+      submit,
+      className,
+      disabled,
+      loading,
+      children
+    } = this.props;
+    return (
+      <button
+        type={submit ? 'submit' : 'button'}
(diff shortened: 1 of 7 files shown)


tajo:
> We need the refs for form components because we are using redux-form on the product.

I think you forgot forms in this PR.

wyuenho:
@tajo thanks, I was rushing to get this PR out without checking git status because, meetings...

PR #163, opened 2017-04-06
Polyfill fetch in Edge browser

sejoker:
reverted changes to logic with Headers for Edge, polyfill fetch for Edge instead


wyuenho:
I like this, but also surprised that global stuff are not frozen.

wyuenho:
Also, the humor of polyfilling the polyfill.

PR #168, opened 2017-04-07
Sync Legacy into Master

sejoker:
- polyfill fetch in Edge
- use `lerna publish --skip-temp-tag`


package.json
```diff
@@ -21,7 +21,7 @@
     "updated": "lerna updated --only-explicit-updates",
     "prettier": "node prettier.js",
     "clean": "lerna clean --yes && rimraf .jest",
-    "publish": "git pull --rebase && check-installed-dependencies && npm run build && lerna publish",
+    "publish": "git pull --rebase && check-installed-dependencies && npm run build && lerna publish --skip-temp-tag",
```
jwineman: I'm fine with the change but don't understand why we added this.  Whats the motivation?

wyuenho: `yarn run publish` keeps failing at the remove dist-tag step for a lot of us in London, no idea why, but this fixed it.

wyuenho:
Hold on to this one. We've discovered an issue with Lerna that whenever we merge a version bump from 1 branch to another, the next time we publish, lerna will detect a change and ask you to bump the version again. If you bump it, and make yet another PR back to the original branch, the version bumping will just get ping pong'ed until eternity. I'd like to figure out a way to deal with this first.

tajo:
@wyuenho @sejoker so what are we going to do about this? this is a serious blocker... we need to keep releasing master

wyuenho:
Okay, let's merge this one last time, and only publish from master for now. We need to have a way to support parallel development, maintenance and publishing of multiple versions of packages. Lerna is currently in the way.

PR #169, opened 2017-04-07
Update README to recommend against external usage

hturan:
There have been a lot of discussions decided recently based to the fact that *"we're the only consumers of cf-ui"*. If this is the case, let's make this an explicit recommendation.

Feel free to suggest alternate copy 😄


jwineman:
I like your phrasing here

koddsson:
Two approvals till a perfect PR.

tajo:
@jculvey I'm jealous I didn't come up with that Archer reference.

PR #172, opened 2017-04-10
Update to React 15.5

tajo:
https://facebook.github.io/react/blog/

Tests are still spitting a bunch of "deprecated" warnings but afaik they come from underlying libs only. I used the codemod but it didn't fix everything.


jculvey:
@tajo Just curious, what didn't work with the codemod? Was it because of inconsistency?

tajo:
@jculvey frankly, I have no idea. It gave me some weird error for 15 files. It also missed one file completely (not giving an error). So far my experience with codemods is not that good. 

Anyway, I think we should merge and release this even if there are many warnings in tests since the ecosystem (libs) needs to move on first. However, we should hold off to update the React version in our applications until all 3rd party apps are fixed.

hturan:
I only just realised this, and it's totally irrelevant, but we can now do:

```js
import {string} from 'prop-types'

const CoolComponent = ({text}) => <marquee>{text}</marquee>

CoolComponent.propTypes = {
  text: string
};
```

💖pretty💖

sejoker:
using 'prop-types' package works fine with previous versions of React.

wyuenho:
This is make the product bundle with 2 separate versions of react. Don't merge this yet.

package.json
```diff
@@ -49,14 +49,16 @@
     "jsdom": "^9.11.0",
     "lerna": "2.0.0-beta.38",
     "prettier": "^0.22.0",
+    "prop-types": "^15.5.6",
     "react": "^15.0.0-0",
     "react-addons-css-transition-group": "^15.0.0-0",
```
wyuenho: This is deprecated, we need to switch to https://github.com/reactjs/react-transition-group instead.

sejoker:
I think we should move `prop-types` into dependencies of each cf-ui package, in that case we won't rely on the consumer. based on [this conversation](https://twitter.com/dan_abramov/status/850537713090658304)

package.json
```diff
@@ -49,14 +49,16 @@
...
-    "react-addons-test-utils": "^15.4.2",
+    "react-addons-test-utils": "^15.5.1",
```
wyuenho: If we only ever test with React 15.5, we should remove this because https://facebook.github.io/react/blog/#react-test-utils

wyuenho:
@sejoker Yes. prop-types is a directly dependency, actually so is React, I think both of them need to go back to dependencies instead of peerDependency. Deduping should no longer be an issue for us, we should only depend on React 15 and up.

wyuenho:
Let's merge this after the fela migration is done.

PR #173, opened 2017-04-11
Add some new constants

tajo:
Added all gradient colors from our styleguide, default boxShadow and main background color.

screen shot 2017-04-10 at 11 54 07 pm

### DIFF
--- a/packages/cf-style-const/src/variables.js
+++ b/packages/cf-style-const/src/variables.js
@@ -4,7 +4,18 @@ const colorOffsetLight = 0.15;
 const colorOffsetDark = 0.07;
 
 export default {
+  gradient: {
+    primary: ['#76C4E2', '#85CBA8'],
+    two: ['#8176B5', '#85CBA8'],
+    three: ['#8176B5', '#76C4E2'],
+    four: ['#8176B5', '#BA77B1'],
+    five: ['#8176B5', '#F16975'],
+    six: ['#F16975', '#F69259'],
+    seven: ['#F69259', '#FFDB6F'],
+    eight: ['#85CBA8', '#FFDB6F']
+  },
   fontSize: '16px',
+  boxShadow: '0 0 20px 0 rgba(136,136,136,0.50)',
   inputFontSize: '13px',
   lineHeight: 1.5,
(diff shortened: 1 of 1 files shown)


packages/cf-style-const/src/variables.js
```diff
@@ -60,5 +71,6 @@ export default {
   colorPurpleLight: lighten(colorOffsetLight, '#d91698'),
   colorPurpleDark: darken(colorOffsetDark, '#d91698'),
   colorTwitterBlue: '#00aced',
-  colorFacebookBlue: '#3b5998'
+  colorFacebookBlue: '#3b5998',
+  mainBackground: '#e6e6e6'
```
hturan: Should this be `colorMainBackground` or `colorLightishGray`? What's our policy on exporting variables for their *purpose*, rather than for what they are?

packages/cf-style-const/src/variables.js
```diff
@@ -4,7 +4,18 @@ const colorOffsetLight = 0.15;
 const colorOffsetDark = 0.07;
 
 export default {
+  gradient: {
```
hturan: If we're providing gradients, let's provide them in a usable format (e.g. `linear-gradient` syntax). Otherwise users can use these in a way that's against design's intent (e.g. `radial-gradient(ellipse farthest-corner at 45px 45px , gradient.primary[0] 0%, gradient.primary[1] 50%)`).

tajo: Good point! Done.

packages/cf-style-const/src/variables.js
tajo: Renamed to `colorMainBackground`. I think we need to combine both approaches since if you want to easily change the backgroundColor, the same constant should not be used for a shadow around inputs.

hturan: Sure thing. It's always the way with naming colour variables, haha. Maybe go with `colorMainBackground: colorLightishGray, colorLightishGray: '#e6e6e6'` or similar? Just that if someone wants `#e6e6e6` they'll probably start using `colorMainBackground` for their shadows, haha.

PR #175, opened 2017-04-12
Move checkBaseTheme from cf-style-const to cf-style-container

tajo:
- `cf-style-container` is a "toolbox" for writing Fela based components, so it's a natural place for this function
- we will be able to remove `cf-style-const` dependency from all our packages except `cf-style-provider`, components should access these consts (variables) through `this.context.theme`, even better through HOCs in `cf-style-provider` like `createComponent()` that passes `this.context.theme` into `props.theme`
- we will be updating `cf-style-const` frequently so lerna prompts updates for all fela components very often (all fela components has `cf-style-const` as dependency)
- `cf-style-container` should be very stable (very few releases) and it should be the only dependency for all Fela components

I hope it makes sense. // @hturan related to my PR review

### DIFF
--- a/packages/cf-component-button/package.json
+++ b/packages/cf-component-button/package.json
@@ -11,7 +11,6 @@
     "registry": "http://registry.npmjs.org/"
   },
   "dependencies": {
-    "cf-style-const": "^0.3.4",
     "cf-style-container": "^0.2.3",
     "polished": "^1.0.2"
   },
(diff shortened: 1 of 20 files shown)


sejoker:
I appreciate your thoughtfulness on that end (care about end users). At this point we should acknowledge - Cloudlfare team is the only consumer of `cf-ui`.  I found it repetitive and unnecessary to use  `checkBaseTheme` function inside cf-ui components. It doesn't provide much value to us. All the use cases are clear from the Readme and Styleguide. I would love to consider removing `checkBaseTheme`  from the codebase.

tajo:
> I found it repetitive and unnecessary to use checkBaseTheme function inside cf-ui components.

Fair enough. I removed it.

PR #178, opened 2017-04-16
added text size bold

sejoker:


packages/cf-component-text/src/Text.js
```diff
@@ -27,7 +27,7 @@ class Text extends React.Component {
 }
 
 Text.propTypes = {
-  size: PropTypes.oneOf(['normal', 'small']),
+  size: PropTypes.oneOf(['normal', 'small', 'bold']),
```
tajo: I think it would be better to add a new prop `style` or `weight` and split `fontSize` and `fontWeight` definitions. It's weird to have `bold` size.

sejoker: made `weight` into separate prop

manatarms:
As per our variable.js bold should be 700. Could we also add a semi-bold to have completeness?  
https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-const/src/variables.js#L25-L28

packages/cf-component-text/src/TextTheme.js
```diff
@@ -1,7 +1,8 @@
 export default baseTheme => ({
   fontSizeNormal: '1em',
-  fontWeightNormal: 400,
   fontSizeSmall: '.8em',
+  fontWeightNormal: 400,
+  fontWeightBold: 600,
```
manatarms: As per our variable.js bold should be 700. Could we also add a semi-bold to have completeness?  
https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-const/src/variables.js#L25-L28

tajo:
👍

packages/cf-component-text/src/TextTheme.js
sejoker: @manatarms @tajo done

PR #180, opened 2017-04-17
Add styleguide colors

tajo:
Adding the style guide colors into the consts. The names were proposed by designers. Every new design should use only these colors. If not push back. This is a fresh start for colors.

screen shot 2017-04-16 at 7 08 45 pm

### DIFF
--- a/packages/cf-style-const/src/variables.js
+++ b/packages/cf-style-const/src/variables.js
@@ -14,6 +14,20 @@ export default {
     seven: 'linear-gradient(left, #F69259, #FFDB6F)',
     eight: 'linear-gradient(left, #85CBA8, #FFDB6F)'
   },
+  color: {
+    rain: '#408BC9',
+    sky: '#76C4E2',
+    dew: '#85CBA8',
+    twilight: '#8176B5',
+    sunset: '#BA77B1',
+    dawn: '#F16975',
+    sunrise: '#F69259',
+    lightning: '#FFDB6F',
+    night: '#404041',
+    dusk: '#4D4D4F',
+    storm: '#808285',
+    hail: '#BCBEC0'
+  },
   fontSize: '16px',
   boxShadow: '0 0 20px 0 rgba(136,136,136,0.50)',
   inputFontSize: '13px',


packages/cf-style-const/src/variables.js
```diff
@@ -14,6 +14,20 @@ export default {
     seven: 'linear-gradient(left, #F69259, #FFDB6F)',
     eight: 'linear-gradient(left, #85CBA8, #FFDB6F)'
   },
+  color: {
+    rain: '#408BC9',
```
manatarms: Are these names final?

jwineman: I would rather rename these something that indicates the color.  We we need to map the design names to our names we should comment `sky` etc next to the hex code.  Even if we do something like `blueSky`, `redDawn` etc.

tajo: These  names were created by designers. @jculvey likes it too. I don't care that much if it is consistent between design and engineering. I'm already using these colors in cf-ux.

jwineman: If design starts defining colors in mocks with these names its probably fine but if they say blue and give us the hexcode for rain we're going to need to rework this.  I think it adds an unnecessary level of confusion in my opinion.

jculvey: Once again opinions and this seem all over the map. 

I like these names but for the sake of time and sanity I'll agree to the first thing other people agree to.

There's 3 ways we can do this:

1. Name them by color. blue1, blue2, blue3 etc or blueDark, blue, blueLight. The advantage here is that there's little mental unpacking on the part of the programmer to figure out what the color names mean. The downside is that it doesn't scale super well if theres a blue lighter than blueLight or blue1.

2. Name them by function. color-primary, color-action-success, etc. We can alias variables as we go along that point at a limited pallette. ex. if button-primary is a blue that gets re-used as a color for info panel borders, we make a new info-panel-border variable that references blue. Advantage is that calling components will be decoupled from actual color names, so renaming is super easy. Downside is there's a lot more variables, and the frequency of new variables adds overhead.

3. Use abstract names like the ones above. Advantage: developers don't think about colors, they just plug in variables indicated in the mocks. There's no naming collisions (like when you have a blue lighter than blueLight or blue1). The downside is that if you want to change all your colors and tomato becomes blue, then you'll need to rename, just like with the first category. Also, if developers really DO want to expend mental energy to decipher the colors, it takes more energy.

jwineman: As long as each hex code is only defined once I don't really care.  For example:
```
color.blue =  '#408BC9';
color.rain = color.blue;
color.success = color.rain;
```

neynah:
So I see that these are new colours proposed, however, how does it affect the current colour palette that we are using? For example, is the proposal that we change all greens to dew: '#85CBA8'? Also, what do we do about different states like hover/disabled?

If we are changing all greens to the dew green for example then we should also be updating the styleguide to reflect all these colour changes (whether in commit or a separate one).

neynah:
Just had a discussion with @tajo, Philina, and @manatarms in person. To clarify, this change is orthogonal to the work @manatarms are doing to standardize the existing colours. As we standardize existing colours, they will be added to this palette that @tajo has started.

blakemattos:
Hi all. Quick process note here: we (the design team) are currently developing our Design approval process on the Design side. We are also drafting Contribution Guidelines which should help us all get on the same page around process. It'll probably look something like these: https://github.com/carbon-design-system/carbon-design-kit/blob/master/CONTRIBUTING.md

Please note these colors are our approved **Brand** palette colors, and are _not_ our approved UI palette. These colors will be changing over time as we define our new palette, so until we have an updated and approved UI palette, we will still using older color palette.

tajo:
@blakemattos Thanks! It's relatively easy to change the names or values of these colors on our end. We just need to keep them in one place - here. I pushed it through fast because I needed them for the onboarding. Feel free to propose changes any time.

PR #182, opened 2017-04-17
pass customTheme optional parameter to override styles from mainTheme

sejoker:
- allows to override styles from the default (main) theme.

### DIFF
--- a/packages/cf-style-container/README.md
+++ b/packages/cf-style-container/README.md
@@ -88,7 +88,7 @@ export default createComponentStyles({ mainStyles, legendStyles }, FormFieldset)
 Notice that rules are now an object. The names you chose will be used for classNames
 accessible as `styles.mainStyles` and `styles.legendStyles` in this case.
 
-### applyTheme(Component, mainTheme, customTheme)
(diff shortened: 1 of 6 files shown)


sejoker:
@tajo I will work on the real integration test in the separate PR.

packages/cf-style-container/README.md
```diff
@@ -88,7 +88,7 @@ export default createComponentStyles({ mainStyles, legendStyles }, FormFieldset)
 Notice that rules are now an object. The names you chose will be used for classNames
 accessible as `styles.mainStyles` and `styles.legendStyles` in this case.
 
-### applyTheme(Component, theme)
+### applyTheme(Component, mainTheme, customTheme)
```
tajo: Could we have even more flexible API using rest parameter syntax like

```js
applyTheme(Component, theme1, theme2 .... theme?)
```

similar to `Object.assign()`. It just seems more idiomatic than having two arguments called `main` and `custom`. What do you think?

sejoker: agreed, I like it  👍 . Ready for the review

PR #188, opened 2017-04-28
card/content: make title optional

koddsson:
Title should be optional

### DIFF
--- a/packages/cf-component-card/src/CardContent.js
+++ b/packages/cf-component-card/src/CardContent.js
@@ -21,7 +21,7 @@ class CardContent extends React.Component {
 }
 
 CardContent.propTypes = {
-  title: PropTypes.any.isRequired,
+  title: PropTypes.node,
   footerMessage: PropTypes.string,
   children: PropTypes.node
 };


wyuenho:
It appears there are "card"-like things in analytics that don't have a title, that I understand. It also appears to be that the cf-ui card may not be entirely appropriate for the cards in analytics. They are mostly tabs, with fairly free-form content, that happen to have drawers. This makes me think whether it's more appropriate to separate the drawers out to a first class package and component, and write a component for analytics more suitable for displaying stat count boxes and charts in www-next, instead of overloading cf-ui Card, which is more suitable for cards that have controls on the side.

jwineman:
If this is for the tiny analytics cards I think it is @thellimist made this:

https://github.com/cloudflare/cloudflare-plugin-frontend/blob/master/src/components/AnalyticCard/AnalyticCard.js

which may or may not be helpful.

koddsson:
@tajo Since abstracting title out as a separate component is going to be a breaking change. I thought we could do it as part of #220. I'll add a comment to that issue.

PR #189, opened 2017-05-04
Modal now uses H1 instead of cf-component-heading.

jwineman:


koddsson:
Sounds good to me. Any particular reason why this change is being made?

jwineman:
Its the result of a discussion we had in a UI architecture meeting a few weeks ago. 

The compromise for deprecating cf-component-heading was to give page, modal, and card their own heading components to make them easier to style and remove the cross component dependency.

koddsson:
Right!

PR #190, opened 2017-05-08
#171 Update cf-component-tooltip to re-render when props change

### LINKED ISSUE #171: cf-component-tooltip doesn't reflect changes in `content` prop
It looks like `cf-component-tooltip` reads the `content` prop once at initialization and never again, meaning that when `content` changes, the changes aren't reflected in the tooltip.

`cf-component-tooltip` should implement `componentDidUpdate` and compare props, then recreate the tooltip if changed.

hturan:
This fixes #171. Currently, the tooltip never updates to reflect new props. With this change, it'll compare props and re-create the tooltip if needed.

### DIFF
--- a/packages/cf-component-tooltip/src/Tooltip.js
+++ b/packages/cf-component-tooltip/src/Tooltip.js
@@ -16,6 +16,20 @@ class Tooltip extends React.Component {
     this.destroyTooltip();
   }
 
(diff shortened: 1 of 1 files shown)


tajo:
You should wait for 2 approvals before merging. ☝️

PR #191, opened 2017-05-11
button: make `onClick` optional

koddsson:
We can render a button without a `onClick` function and this patch will make sure we don't get prop types errors while developing.

### DIFF
--- a/packages/cf-component-button/src/Button.js
+++ b/packages/cf-component-button/src/Button.js
@@ -148,7 +148,7 @@ class Button extends React.Component {
 }
 
 Button.propTypes = {
-  onClick: PropTypes.func.isRequired,
+  onClick: PropTypes.func,
   submit: PropTypes.bool,
   spaced: PropTypes.bool,
   className: PropTypes.string.isRequired,


wyuenho:
It's okay to see prop errors while developing. It reminds you that eventually you'll have to put an onClick handler back. That's the whole point of prop types.

hturan:
We have at least 3 cases in the internal codebase where we set `onClick={() => {}}` due to this restriction. Sometimes we want a button that has no function to keep a UI stable in certain situations.

wyuenho:
If you want to pass `()=>{}` to quiet down prop types, that's a conscious decision, it's explicit. You look at it and wonder "huh, this is wrong, I've gotta fix this." `()=>{}` serves the same purpose as a warning on the console, just less noisy.

2 instances in the code base are stubs for WIP code, the one for rate limiting is the only instance where we consciously want a non-disbaled button that does not trigger any action. It's useful to have some mechanism to remind you possible errors while not stopping you from getting around it. Making it optional take this help away.

wyuenho:
Yes you can. The thing is, setting the prop type as required doesn't prevent you from doing anything. If you know what you are doing, you can make a conscious choice to get around it. I do want to force that explicit behavior, and leave the code as is to help the people who want it. Explicit is often better than implicit because of clarity.

Issue #195, opened 2017-05-17
Almost all tests are broken


koddsson:
### How to reproduce:

1. clone project
2. `npm run bootstrap`
3. `npm run build`
4. `npm run test`

### Expected behaviour: 

The tests run and come back passing

### Actual behaviour:

<details>
<summary>The tests run and all but one are failing.</summary>

```
...

 FAIL  packages/cf-component-modal/test/ModalBody.js
  ● Test suite failed to run

    /Users/koddsson/Repos/cf-ui/packages/cf-component-modal/test/ModalBody.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import React from 'react';
                                                                                             ^^^^^^
    SyntaxError: Unexpected token import

      at transformAndBuildScript (node_modules/jest-runtime/build/transform.js:320:12)
      at handle (node_modules/worker-farm/lib/child/index.js:41:8)
      at process.<anonymous> (node_modules/worker-farm/lib/child/index.js:47:3)
      at emitTwo (events.js:106:13)

 FAIL  packages/cf-builder-table/test/TableBuilder.js
  ● Test suite failed to run

    /Users/koddsson/Repos/cf-ui/packages/cf-builder-table/src/TableBuilder.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){var _createClass = function () {function defineProperties(target, props) {for (var i = 0; i < props.length; i++) {var descriptor = props[i];descriptor.enumerable = descriptor.enumerable || false;descriptor.configurable = true;if ("value" in descriptor) descriptor.writable = true;Object.defineProperty(target, descriptor.key, descriptor);}}return function (Constructor, protoProps, staticProps) {if (protoProps) defineProperties(Constructor.prototype, protoProps);if (staticProps) defineProperties(Constructor, staticProps);return Constructor;};}();function _classCallCheck(instance, Constructor) {if (!(instance instanceof Constructor)) {throw new TypeError("Cannot call a class as a function");}}function _possibleConstructorReturn(self, call) {if (!self) {throw new ReferenceError("this hasn't been initialised - super() hasn't been called");}return call
    SyntaxError: Unexpected token import

      at transformAndBuildScript (node_modules/jest-runtime/build/transform.js:320:12)
      at Object.<anonymous> (packages/cf-builder-table/src/index.js:3:21)
[...]

PR #196, opened 2017-05-18
deps: update jest

koddsson:
Seems to not make all the tests fail :laughing:

Fixes: #195


koddsson:
To be clear there are some tests failing tho but they seem to be valid failures.

koddsson:
Nope. Didn't really look into it at all.

Issue #100, opened 2017-03-16
Fela Migration


jwineman:
Lets track the migration of these components here.  Call out what you're working on and I'll keep this thread up to date.

## 8 / 33 ( 24% ) complete.

- [x] cf-component-button @tajo https://github.com/cloudflare/cf-ui/pull/77		
- [ ] cf-component-callout		
- [ ] cf-component-card		
- [ ] cf-component-checkbox		
- [ ] cf-component-code		
- [ ] cf-component-copyable-textarea	
- [ ] cf-component-dropdown		
- [ ] cf-component-dynamic-content		
- [ ] cf-component-flex		
- [x] cf-component-form @tajo https://github.com/cloudflare/cf-ui/pull/133
- [x] cf-component-heading @sejoker https://github.com/cloudflare/cf-ui/pull/97
- [ ] cf-component-icon @koddsson	
- [ ] cf-component-input		
- [x] cf-component-label @manatarms 		
- [ ] cf-component-layout		
- [x] cf-component-link	@manatarms 
- [ ] cf-component-list		
- [ ] cf-component-loading		
- [ ] cf-component-modal		
- [ ] cf-component-notifications		
- [x] cf-component-page @manatarms https://github.com/cloudflare/cf-ui/pull/102		
- [x] cf-component-pagination @sejoker https://github.com/cloudflare/cf-ui/pull/135	
- [ ] cf-component-progress		
- [ ] cf-component-radio		
- [ ] cf-component-select
- [ ] cf-component-table
- [ ] cf-component-tabs		
- [x] cf-component-text @sejoker https://github.com/cloudflare/cf-ui/pull/98		
- [ ] cf-component-textarea		
- [ ] cf-component-toggle		
- [ ] cf-component-tooltip		
- [ ] cf-component-typography		
- [ ] cf-component-viewport

manatarms:
cf-component-page please!

wyuenho:
Please stop merging into master for now. We agree on how to support a "stable" version of cf-ui and work on big changes like this.

manatarms:
cf-component-label in PR #103

tajo:
Afaik, it should be fine to merge things. Just don't forget to major bump the new components when you are publishing them for the first time. We can still do bug fix releases of older versions.

tajo:
Please check out this on our wiki: WWW / JavaScripts / How to update cf-ui components to Fela

tajo:
## Note

[...]

PR #199, opened 2017-05-19
icon: migrate cf-component-icon to fela

koddsson:
I made a bunch of assumptions in here that you can call me out on. Figured I should plow on with what I thought would be best.

Related-to: #100


packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-class Icon extends React.Component {
-  render() {
-    let className = `cf-icon cf-icon--${this.props.type}`;
+import { createComponent } from '../../cf-style-container';
+
+const Wrapper = createComponent(({
+  color = '#333333',
```
koddsson: Should I be taking this in from theme somehow?

koddsson: somehow === `theme.color`

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
...
-      className += ` cf-icon--${this.props.size}`;
+  if (spin) {
+    additionalRules.display = 'inline-block';
+    additionalRules.animation = `${theme.spinAnimationName} 2s infinite linear`;
+  }
+
+  return {
+    color,
+    fontSize,
+    fontStyle: 'normal',
```
koddsson: We need this to override the italics of `<i />`.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
...
   }
 }
 
 Icon.propTypes = {
-  label: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.oneOf([false])
-  ]).isRequired,
+  /**
+   * The a11y label set for screen-readers(?). Needs to correspond with the type
```
koddsson: Is this correct?

packages/cf-component-icon/src/Icon.js
manatarms: If I remember correctly it should be in cf-style-const?

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
...
     }
+  };
+}, 'i');
 
-    if (this.props.border) className += ' cf-icon--border';
-    if (this.props.spin) className += ' cf-icon--spin';
-    if (this.props.muted) className += ' cf-icon--muted';
-    if (this.props.white) className += ' cf-icon--white';
+const Label = createComponent(() => ({
+  display: 'none'
```
koddsson: If this is or `a11y` do we need to hide this differently so screen readers can 'see' it?

packages/cf-component-icon/src/Icon.js
```diff
@@ -76,11 +288,18 @@ Icon.propTypes = {
...
+   */
   border: PropTypes.bool,
+  /**
+   * Renders icon with spinning animation
+   */
   spin: PropTypes.bool,
-  muted: PropTypes.bool,
-  white: PropTypes.bool,
+  /**
+   * TODO: What is this?
```
koddsson: Wat is?

packages/cf-component-icon/test/__snapshots__/Icon.js.snap
```diff
@@ -1,54 +1,37 @@
...
-/>
-`;
-
 exports[`should render type 1`] = `
 <i
-  className="cf-icon cf-icon--info-sign"
-  role={undefined}
+  className="cf-1ew9cp5"
+  id={undefined}
+  style={undefined}
```
koddsson: I'm not setting these props so I'm not sure why these are here, maybe something with `createComponent`?

packages/cf-style-provider/src/index.js
```diff
@@ -36,11 +38,22 @@ export const createRenderer = opts => {
...
+  renderer.renderFont('Cloudflare Icons', [
+    `data:application/x-font-ttf;charset=utf-8;base64,${cloudflareIcons}`
+  ]);
+
+  const spinAnimationName = renderer.renderKeyframe(() => ({
+    from: { transform: 'rotate(0deg)' },
+    to: { transform: 'rotate(359deg)' }
+  }));
+
+  return { renderer, spinAnimationName };
```
koddsson: Not sure if there's a cleaner way to do this.

packages/cf-style-provider/src/index.js
```diff
@@ -18,6 +18,8 @@ import { Provider } from 'react-fela';
 import { variables } from 'cf-style-const';
 import { ThemeProvider } from 'cf-style-container';
 
+import cloudflareIcons from './cloudflare-icons';
```
koddsson: Not sure if there is a cleaner way to do this.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
...
+    case 'close':
+      fontPoint = "'\\D7'";
+      additionalRules.fontFamily = 'Open Sans, Helvetica, Arial, sans-serif';
+      break;
+  }
+
+  if (border) {
+    additionalRules.border = `1px solid ${color}`;
+    additionalRules.padding = '.2em .25em .15em';
+    additionalRules.borderRadius = '2px';
```
manatarms: The consts also have font-family,border radius and padding.

packages/cf-component-icon/test/__snapshots__/Icon.js.snap
```diff
@@ -1,54 +1,37 @@
...
-  role="status"
-/>
-`;
-
 exports[`should render type 1`] = `
 <i
-  className="cf-icon cf-icon--info-sign"
-  role={undefined}
+  className="cf-1ew9cp5"
+  id={undefined}
```
manatarms: Are these supposed to be undefined?

packages/cf-component-icon/src/Icon.js
koddsson: Sounds good. What if this padding is specific to this component? Should we still abstract it out to a constant to be used (probably) in one place?

packages/cf-component-icon/test/__snapshots__/Icon.js.snap
koddsson: I made a comment about it elsewhere in this PR. Seems to come from `createComponent`. I'm at least not adding these props anywhere.

packages/cf-component-icon/src/Icon.js
manatarms: If you're only using it once it should be fine. The constants in cf-style-conts are the base theme that would generally apply to most components. Also it's a good idea to break your padding out into atomic styles for Fela. So its better to have padding-top,bottom etc

koddsson: Yeah so might keep this in here but break it up.

packages/cf-component-icon/src/Icon.js
hturan: 😵

hturan:
I love that CI chose to fail on this, but not on anything else.

packages/cf-component-icon/src/Icon.js
hturan: Why are we setting this regardless?

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
...
-    let className = `cf-icon cf-icon--${this.props.type}`;
+import { createComponent } from '../../cf-style-container';
+
+const Wrapper = createComponent(({
+  color = '#333333',
+  fontSize = 'initial',
+  border = false,
+  spin = false,
+  type,
+  theme
```
hturan: Lets add in the removed `color`, `muted` and `size` props too.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
...
 Icon.propTypes = {
-  label: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.oneOf([false])
-  ]).isRequired,
+  /**
+   * The a11y label set for screen-readers(?). Needs to correspond with the type
+   * prop.
+   */
+  label: PropTypes.oneOf([
```
hturan: I think the `label` prop should be left as `PropTypes.string`, since we can provide additional context (e.g. "open menu", rather than just "hamburger").

packages/cf-component-icon/src/Icon.js
koddsson: It's set in the css rule in cf-www-next.

packages/cf-component-icon/src/Icon.js
koddsson: I was hoping to get rid of those with this change :P seemed to be just changing the color and you should be able to do it yourself with theming.

koddsson:
After some chatting with @hturan we're realizing that this is missing a theme so I'm going to have to add that.

packages/cf-component-icon/src/Icon.js
hturan: Yeah, makes sense. If we could customise `color` that wold solve all issues.

packages/cf-component-icon/src/Icon.js
hturan: Hm, I can't see it in `icon.css`.

packages/cf-component-icon/test/__snapshots__/Icon.js.snap
tajo: Yea, Fela did it. Ignore that.

hturan:
> After some chatting with @hturan we're realizing that this is missing a theme so I'm going to have to add that.

Maybe! I'm not 100% sure how to customise `fela` components. It looks like `Wrapper` already accepts a theme?

packages/cf-component-icon/src/Icon.js
tajo: Don't import cf-style-const into cf-components. Always use props.theme.color.

koddsson:
>> After some chatting with @hturan we're realizing that this is missing a theme so I'm going to have to add that.

>Maybe! I'm not 100% sure how to customise `fela` components. It looks like `Wrapper` already accepts a theme?

I'm taking a stab at this right now, incoming commit.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,33 +1,245 @@
...
+  fontSize = 'initial',
+  border = false,
+  spin = false,
+  type,
+  theme
+}) => {
+  const additionalRules = {};
+  let fontPoint = '';
+
+  switch (type) {
```
tajo: It's missing default case. Also, if you turn this switch into function, you can save a lot of lines. :)

packages/cf-component-icon/src/Icon.js
```diff
@@ -76,11 +288,18 @@ Icon.propTypes = {
     'widen',
     'wrench'
   ]).isRequired,
-  size: PropTypes.oneOf(['2x', '3x', '4x', 'large', 'xlarge']),
```
tajo: Why don't we have size prop anymore?

packages/cf-style-provider/src/index.js
tajo: Yea, there is now: https://github.com/rofrischmann/fela/tree/master/packages/fela-plugin-embedded
It was originally my [proposal](https://github.com/rofrischmann/fela/pull/238) and it just landed.

**Please, wait with this PR until the new version of Fela is released and we'll use embedded plugin.**

tajo:
@koddsson Ad theming: every cf-ui component should have its own theme. Look at other transitioned components. The final component look is a combination of the base theme (cf-style-const) that's already included in cf-www-next through the context and `createComponent()` assigns the base theme from context to the prop `theme`. The second piece is component's own theme.  We export our components themed already (`applyTheme()`) but there is an escape hatch if you really want to overwrite defaults - you can do something like:

```js
import { ButtonUnstyled, ButtonTheme } from 'cf-component-button';
import { applyTheme } from 'cf-style-container';

const MyButton = applyTheme(ButtonUnstyled, ButtonTheme, { prop_i_want_to_overwrite: 'value' })
```

So more stuff you abstract into the IconTheme means more things you can customize.

packages/cf-component-icon/test/__snapshots__/Icon.js.snap
rofrischmann: We can add a check and only add it if it actually has a value if that'd make sense :)

packages/cf-component-icon/src/Icon.js
koddsson: I kind of want this to error if there is no type. I think I might either do a check in the very beginning to `type` and throw an error there or in the `default` case. Left it open since I wasn't sure at the time of writing, will change this soon.

packages/cf-component-icon/src/Icon.js
koddsson: Cool! I wanted that too but was worried that label needed to be something specific. Since no one has made any additional comments I think we should def do that. Will change this.

packages/cf-component-icon/src/Icon.js
koddsson: Ah, I took it out since I figured people would just use the `fontSize` CSS attribute at this point.

packages/cf-component-icon/src/Icon.js
koddsson: Oh yeah the default way is def the way to go since otherwise I'd have to have a very long if statement 😄

koddsson: Or maybe it's not needed since the `type` proptype is set to being a `required` prop.

packages/cf-component-icon/src/Icon.js
hturan: We probably want to limit the sizes of icons for consistency. Worth checking with design.

koddsson: We need a @cloudflare/design!

packages/cf-builder-pagination/src/PaginationBuilder.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Pagination, PaginationItem } from 'cf-component-pagination';
-import Icon from 'cf-component-icon';
+import Icon from '../../cf-component-icon/src';
```
koddsson: This is probably wrong since this will break these components. Wondering what the best solution here would be. Probably a webpack/babel alias?

koddsson:
My local env is still throwing errors when ever which is why there are so many commits 😓

packages/cf-component-icon/test/__snapshots__/Icon.js.snap
koddsson: @rofrischmann Yeah I think there's value in that. It would make these snapshot comparisons less noisy and easier to spot if one of those props are added or removed.

koddsson:
@hturan @manatarms Time to take another look at this for me?

packages/cf-builder-pagination/src/PaginationBuilder.js
hturan: Hm, yeah, this is interesting. Is the only reason we need this to make tests pass? Can't we reference a specific version of `cf-component-icon` in `cf-builder-pagination`?

koddsson: If we could the issue would be that the new version isn't on `npm` yet

koddsson: If I'm understanding correctly

packages/cf-component-icon/src/Icon.js
tajo: So our new icons might have different level of details for different sizes, so it makes sense to have this under the control.

koddsson: Ok I'll put it back but I think there's value in trying to get rid of code not needed anymore and maybe not keep un-used code around just in case we might need it in the future.

packages/cf-builder-pagination/src/PaginationBuilder.js
tajo: I think you just forgot to run `npm run build` before `npm test`. Stuff is already symlinked by lerna but it's pointing to `packages/cf-component-icon/lib/index.js`, so you need to rebuild before running tests.

> This is probably wrong since this will break these components.

Yep.

koddsson: Hmm. I think I added this because CI was failing. I'll remove all these and make a commit with it and hopefully it will just work :)

packages/cf-component-icon/src/Icon.js
hturan: FWIW, we're using `size="large"` in the www-next codebase, but it's only in one place.

koddsson: Maybe that's a signal to us to remove this since it's basically unused.

packages/cf-component-icon/src/Icon.js
koddsson: What is a `role`? Should I remove it?

manatarms: I think it might be aria stuff. I cant see the file anymore so I don't know for sure.

koddsson:
I'd like to get this in so I can build on top of it. How does the release schedule work for fela? @tajo @rofrischmann ?

rofrischmann:
Was about to release yesterday, but found a bug which needed to be fixed first.. Maybe I get ready today, but right now I am sitting in a car to vienna :p I might add @tajo to the npm package(s) for further releases

koddsson:
No worries, there's no rush. Just pondering :)

manatarms:
Can we add the new icons in here too? Or do you want to separate that out into a different PR.

koddsson:
@manatarms Seperate after this one. I don't want this to grow into a unmergable behemoth.

tajo:
@koddsson is this blocking you? It actually can take more time to release another version of fela (I would say at least a week or two).

koddsson:
@tajo Nope not currently blocking, I just thought I'd help move the needle a bit on the fela migration. Only worry is that this is going to get stale but I think this can wait for a week or two so we can do this right :)

manatarms:
Is there an eta on when this will get merged? I want to make a PR to add the new icon codes.

tajo:
@manatarms a week max

manatarms:
Any updates on this? I really want to add the new icon types to the prop types.

tajo:
https://github.com/cloudflare/cf-ui/pull/248

koddsson:
I'm updating this now!

koddsson:
@tajo Can you point me on how I'm gonna achieve this with the embedd plugin? How will I access the keyframe name?

koddsson:
I maybe got it by reading the source. Does it add the `animationName` and `fontFace` properties passed to it into the `baseTheme`?

manatarms:
I think in order to make this PR complete we should also update the propTypes to allow the new icons. This ensures that when this goes out it'll be synced with www-next
Here's the list if you decide to do it.

```
.cf-icon--caret-up::before,
.cf-icon--chevron-up::before        { content: "\f000"; }
.cf-icon--caret-down::before,
.cf-icon--chevron-down::before      { content: "\f001"; }
.cf-icon--caret-left::before,
.cf-icon--chevron-left::before      { content: "\f002"; }
.cf-icon--caret-right::before,
.cf-icon--chevron-right::before     { content: "\f003"; }
.cf-icon--ok-sign::before           { content: "\f004"; }
.cf-icon--exclamation-sign::before  { content: "\f005"; }
.cf-icon--info-sign::before         { content: "\f006"; }
.cf-icon--pause::before             { content: "\f007"; }
.cf-icon--time::before              { content: "\f008"; }
.cf-icon--ok::before                { content: "\f009"; }
.cf-icon--clipboard::before         { content: "\f00a"; }
.cf-icon--bolt::before              { content: "\f00b"; }
.cf-icon--chart::before             { content: "\f00c"; }
.cf-icon--drive::before             { content: "\f00d"; }
.cf-icon--filter::before            { content: "\f00e"; }
.cf-icon--flowchart::before         { content: "\f00f"; }
.cf-icon--hamburger::before         { content: "\f010"; }
.cf-icon--lock::before              { content: "\f011"; }
.cf-icon--list::before              { content: "\f012"; }
.cf-icon--file::before              { content: "\f013"; }
.cf-icon--shield::before            { content: "\f014"; }
.cf-icon--plus::before              { content: "\f015"; }
.cf-icon--happy::before             { content: "\f016"; }
.cf-icon--sad::before               { content: "\f017"; }
.cf-icon--search::before            { content: "\f018"; }
.cf-icon--wrench::before            { content: "\f019"; }
.cf-icon--remove::before,
.cf-icon--remove-sign::before       { content: "\f01a"; }
.cf-icon--speech::before            { content: "\f01b"; }
.cf-icon--gear::before              { content: "\f01c"; }
.cf-icon--help::before              { content: "\f01d"; }
.cf-icon--calendar::before          { content: "\f01e"; }
.cf-icon--facebook::before          { content: "\f01f"; }
.cf-icon--twitter::before           { content: "\f020"; }
.cf-icon--googleplus::before        { content: "\f021"; }
.cf-icon--linkedin::before          { content: "\f022"; }
.cf-icon--upload::before            { content: "\f023"; }
.cf-icon--widen::before,
.cf-icon--resize-horizontal::before { content: "\f024"; }
.cf-icon--resize-vertical::before,
.cf-icon--reorder::before           { content: "\f024"; transform: rotate(90deg); }
.cf-icon--download:before           { content: "\e900"; }
.cf-icon--refresh:before            { content: "\e901"; }
.cf-icon--api:before                { content: "\e902"; }
.cf-icon--firebolt:before           { content: "\e903"; }
.cf-icon--mail:before               { content: "\e904"; }
.cf-icon--network:before            { content: "\e905"; }
.cf-icon--door:before               { content: "\e906"; }
.cf-icon--credit-card:before        { content: "\e93f"; }
```

packages/cf-component-icon/src/Icon.js
wyuenho: These styles came out of the existing backbone code base, they could be useful at some point. If it's not much effort, I think it makes sense to leave that in.

packages/cf-style-provider/src/index.js
wyuenho: Actually, I'm not sure why embedding is even necessary or desired. How do we share the same font file from the backbone side and Fela side without sharing the same font URL? What about caching?

packages/cf-component-icon/src/Icon.js
wyuenho: You can still provide a type that doesn't exist. If you extract this giant switch block into a function, you may be able to provide it as a custom proptype.

koddsson:
>Can you point me on how I'm gonna achieve this with the embedd plugin? How will I access the keyframe name?

I actually figured this out. Working on this now.

packages/cf-component-icon/src/Icon.js
koddsson: So I was hoping we'd leverage the power of `fela` to set the size in the context of how the icon is used? To me that makes more sense so I thought I'd deprecate it

koddsson:
Ok I'm lost again. Could you point me to how I can use the embedded plugin with a base64 encoded font @tajo? 

The docs seem to suggest that I can but there's a check for `src` to be an array in the embedded plugin which I don't get as an array with a base64 font but rather a string and I'm hitting walls in trying to debug this in fela.

koddsson:
Ok I've smoke tested the component using `cf-styleguide` and it is working as intended from what I can see.

Here's a list of things I've done in this PR some of which are subject to debate:

- Migrated the Icon component to use the CSS-in-JS framework `fela` instead of relying on external styles.
- Shipped the icon fonts with the Icon component as base64 encoded string
- Deprecated a lot of properties. `muted`, `border`, `size` and `role`.
  - I figured people could leverage `fela` to add those styles where applicable.
- Updated all the icon references from the updated list from @manatarms.

I really think that this is the time to refactor and deprecate things we aren't using to keep our components lean and simple so I opted for ripping things out that I can't see being used all that much.

wyuenho:
> Deprecated a lot of properties. muted, border, size and role.
> I figured people could leverage fela to add those styles where applicable.

Size defines a list of precalculated sizes so the user doesn't have to. The size and color are generally the most commonly changed style properties in icon fonts, although this hasn't been the case at Cloudflare. The reason for that is, we allow the typographic styles to cascade into the icons.

The way you are isolating the color and font-size in the icon component defeats the purpose of icon fonts. Icon fonts are meant to use the cascaded font color and sizes by default, and allow in-place override as a property, hence `size`.

However, I think it's probably worthwhile to get @mrmrs to review whether the predefined sizes still make sense.

`role` is also an ARIA accessible attribute. Since we are not yet allowing arbitrary valid HTML props to pass down to dumb components (see the whitelist vs blacklist debate), let's not change the API too much now.

koddsson:
I'd love to try to do the `role` property by allowing all the props instead in all components, it would allow devs to use the `aria-*` attributes as well.

An alternative would be defining the `role` and `aria-*` attributes and [their values](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) in a seperate file and import those in every component to inject into the prop types.

koddsson:
I haven't heard anything back so I'm going to assume that people want the API back to how it was. Working on that now.

wyuenho:
Would love to preserve the ability to do this:

```jsx
const p = createComponent('p', {color: 'green'});
<p>
  <Icon type="ok" /> // rendered a green span of text, no need to use applyTheme to set a color
</p>
```

koddsson:
@wyuenho I can add a `color` prop if you want but I think cascading styles like that aren't possible with `fela`(?)

wyuenho:
It is possible. Fela doesn't "isolate" the styles of the component, despite commonly believe. CSS just cascades, the only way to stop that from happening is to redefine all the CSS properties on all the HTML elements or use [all: initial](https://developer.mozilla.org/en/docs/Web/CSS/all), which also means you have to redefine some CSS properties on all the HTML elements for typographic styles.

koddsson:
Then can we do the `font-size` similarly like I suggested earlier in this PR?

wyuenho:
Try this:
```jsx
const div = createComponent({text-align: 'left'});
<div>
  <p>
    hello world
  </p>
</div>
```
Guess which direction the text inside p is aligned?

wyuenho:
@koddsson Yes, but make sure the `color` property is filtered out from the wrapper when if one isn't supplied as a prop

koddsson:
I'm super confused. Could you write up the props you would like this component to accept? It feels from these last comments that you don't want `size` or `color` anymore and it should cascade from the parent? Is that correct?

koddsson:
Waiting on @mrmrs to come in and do stuff

packages/cf-component-icon/README.md
```diff
@@ -45,57 +55,34 @@ const IconComponent = () => (
...
       <Icon label="remove" type="remove" />
       <Icon label="remove-sign" type="remove-sign" />
+      <Icon label="reorder" type="reorder" />
       <Icon label="resize-horizontal" type="resize-horizontal" />
+      <Icon label="resize-vertical" type="resize-vertical" />
       <Icon label="sad" type="sad" />
       <Icon label="search" type="search" />
       <Icon label="shield" type="shield" />
-      <Icon label="speech" type="speech" />
+      <Icon label="sppech" type="speech" />
```
manatarms: Typo here :)

packages/cf-component-icon/example/basic/component.js
```diff
@@ -32,47 +42,45 @@ const IconComponent = () => (
...
       <Icon label="remove" type="remove" />
       <Icon label="remove-sign" type="remove-sign" />
+      <Icon label="reorder" type="reorder" />
       <Icon label="resize-horizontal" type="resize-horizontal" />
+      <Icon label="resize-vertical" type="resize-vertical" />
       <Icon label="sad" type="sad" />
       <Icon label="search" type="search" />
       <Icon label="shield" type="shield" />
-      <Icon label="speech" type="speech" />
+      <Icon label="sppech" type="speech" />
```
manatarms: And here same.

packages/cf-component-icon/example/basic/component.js
```diff
@@ -32,47 +42,45 @@ const IconComponent = () => (
...
-
-    <p>Muted:</p>
-    <p><Icon label="border" type="gear" muted /></p>
+    <p><Icon label="gear" type="gear" spin /></p>
 
-    <p>White:</p>
-    <p><Icon label="border" type="gear" white /></p>
+    <p>Sizes (hover for the size prop value):</p>
+    <p>
+      <Icon label="1.5x" type="gear" size="1.5x" />
```
manatarms: Thank you for this. I really appreciate it.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,32 +1,178 @@
...
+
+class Icon extends React.Component {
+  render() {
     return (
-      <i className={className} role={this.props.role}>
-        {this.props.label &&
-          <span className="cf-icon__label">{this.props.label}</span>}
-      </i>
+      <Wrapper {...this.props} title={this.props.label}>
+        <Label>{this.props.label}</Label>
```
wyuenho: The label wasn't meant to the title of whatever wrapper element. It was literally meant to be the associated text of the icon.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,32 +1,178 @@
...
+);
+
+class Icon extends React.Component {
+  render() {
     return (
-      <i className={className} role={this.props.role}>
-        {this.props.label &&
-          <span className="cf-icon__label">{this.props.label}</span>}
-      </i>
+      <Wrapper {...this.props} title={this.props.label}>
```
wyuenho: This spreads the Icon props into the wrapping `<i>` element too, you'll get a warning from React for that. Please filter out the Icon props first before you pass them down.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,32 +1,178 @@
...
+        animationDuration: '2s',
+        animationIterationCount: 'infinite',
+        animationTimingFunction: 'linear'
+      });
+    }
+
+    const { content, ...rest } = iconData;
+
+    return {
+      color: theme.color,
```
wyuenho: Take `color` out of the theme. Don't take a default from the theme, take it from `props`. So the user can override the color by setting on `Icon` prop directly. If not set, just keep this `undefined`, so the color can cascade from the parent.

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,32 +1,178 @@
...
+      fontStyle: theme.fontStyle,
```
wyuenho: Deal with this like `color`. Remove it from the theme, open a prop, default this to undefined.

.babelrc
```diff
@@ -1,6 +1,6 @@
 {
   "presets": [
-    ["es2015", { "modules": false }],
```
sejoker: with `{ modules: false}` we define not to transform modules to commonjs. https://babeljs.io/docs/plugins/preset-es2015/#optionsmodules. What was the motivation to change to commonjs?

packages/cf-component-icon/example/basic/component.js
```diff
@@ -3,8 +3,15 @@ import Icon from 'cf-component-icon';
 
 const IconComponent = () => (
   <div>
-    <p>Types:</p>
-    <p>
+    <p>Types (hover for type prop value):</p>
+    <div
+      style={{
+        display: 'grid',
```
sejoker: I want to clarify, should we promote use of grid layout with IE/Edge support old format and no support in Safari 9 https://caniuse.com/#search=grid?

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,32 +1,178 @@
...
+      fontSize: theme.fontSize,
+      ...rest,
+      fontFace: {
+        fontFamily: 'Cloudflare Icons',
+        src: [`data:application/x-font-ttf;base64,${cloudflareIcons}`]
```
sejoker: more general question: are we going to deprecate Cloudlfare icons font instead of svg? We brought this question earlier but I don't remember the final decision we made. https://css-tricks.com/icon-fonts-vs-svg/

.babelrc
tajo: +1, by default we don't want to compile ES modules into commonjs, so tooling as webpack can do tree shaking. For testing & node.js there's `commonjs` babel env adding that.

packages/cf-component-icon/src/Icon.js
manatarms: I'm working on estimating the effort for the switch to svgs. I'd rather have the Fela icon component stabilize first before making big changes.

packages/cf-component-icon/example/basic/component.js
koddsson: This was just in the example so I figured it would be ok. I feel like we don't need to support as many browsers in there.

tajo:
@koddsson I got lost in this PR. Can you please summarize why it is blocked? What needs to be done?

koddsson:
Sorry https://github.com/cloudflare/cf-ui/pull/199#issuecomment-310434695 wasn't very helpful.

@mrmrs asked me to hold off for now since he has ideas and thoughts about how icons should be. SVGs and whatnot instead of font icons.

manatarms:
Closing in favor of SVG implementation. I guess this makes it a total of 107 comments.

PR #201, opened 2017-05-22
use pagination background color from the new design colors

sejoker:
screen shot 2017-05-22 at 9 09 56 pm


manatarms:
@sejoker Snapshots need update

koddsson:
Actually maybe my use case should just be a different theme?

sejoker:
@koddsson make sense to me. In my case I need only `dotted` style. Not aware of our usages of `cf-component-pagination` in the dashboard

koddsson:
>Not aware of our usages of cf-component-pagination in the dashboard

Yup. We're adding the first one right now and it looks understandably out of place. Jut added #204 to add the theme and working on that now.

PR #203, opened 2017-05-24
introduce new light gray color dessy

sejoker:
This is a request from the design team to introduce new light gray color.
screen shot 2017-05-24 at 10 12 12 am

### DIFF
--- a/packages/cf-style-const/src/variables.js
+++ b/packages/cf-style-const/src/variables.js
@@ -33,8 +33,7 @@ export default {
     grass: '#9BCA3E',
     marine: '#2F7BBF',
     tangerine: '#FF7900',
-    apple: '#BD2527',
-    dessy: '#F7F7F7'
+    apple: '#BD2527'
   },
   fontSize: '15px',
   boxShadow: '0 0 20px 0 rgba(136,136,136,0.50)',
@@ -75,6 +74,7 @@ export default {
   colorOrangeDark: darken(colorOffsetDark, '#ff7900'),
   colorGray: darken(0.5125, '#fff'),
   colorGrayLight: darken(0.13, '#fff'),
+  colorGrayLightOnboarding: '#F7F7F7',
   colorGrayDark: darken(0.8, '#fff'),
   colorGrayBorder: '#666',
   colorGreen: '#9bca3e',
--- a/packages/cf-style-const/src/variables.js
+++ b/packages/cf-style-const/src/variables.js
@@ -33,7 +33,8 @@ export default {
     grass: '#9BCA3E',
     marine: '#2F7BBF',
     tangerine: '#FF7900',
-    apple: '#BD2527'
+    apple: '#BD2527',
+    dessy: '#F7F7F7'
   },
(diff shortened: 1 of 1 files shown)


koddsson:
Do these color names make sense? Will design label all the colors on the designs so we know which one is which?

packages/cf-style-const/src/variables.js
```diff
@@ -33,7 +33,8 @@ export default {
     grass: '#9BCA3E',
     marine: '#2F7BBF',
     tangerine: '#FF7900',
-    apple: '#BD2527'
+    apple: '#BD2527',
+    dessy: '#F7F7F7'
```
koddsson: Can we call this color `lightGray` if it's light gray?

koddsson:
This big red x in github feels so agressive 😭

hturan:
Yeah, the only way named colours ever work is if we're consistent with labelling and we document what they look like. We do neither at the moment, so using them is super hard.

packages/cf-style-const/src/variables.js
koddsson: Mmm.. `colorGrayLighter`? 😄 I'm really not sure what to call it but just maybe connect it more to what the value is exactly.

Otherwise I'd be happy to just have a very simple color pallet of our most frequently used colors and just hardcode the one-offs (this feels like a one off) in the themes of the components using it. If it's not gonna be used as the light gray color everywhere then it probably should be localized to where it's used.

koddsson:
>It's an easy problem to solve though. Displaying these in the cf-ui documentation should be trivial.

I'm thinking maybe also when we get designs internally, like knowing the difference between this gray color and that gray color. Like is it `fog` or `dessy` or what?

Maybe I'm overthinking this.

If there's just the gray colors with `dark` and `light` pre/post fixes it's easy enough to kind of find it but cycling through all the names of colors just to find the correct gray color is maybe a bit too much.

sejoker:
@koddsson all this color naming mess is confusing to me as well, hope the new designer who will work on the company's style guide will put this into order.

sejoker:
@koddsson I renamed the color as was discussed.

manatarms:
I'm working on the colors with Nena. Hopefully it should all make some sense soon.

neynah:
Obviously we should name the colours in a way that makes implementation easier but we haven't seemed to land on anything. The sane thing to do in the interim seems to be to just name them consistently at least.

@sejoker Totally. As it stands, the designers are indifferent about how we name colours but I'll discuss this with our new designer who will be working on style guide and see if he's got any ideas.

sejoker:
I am merging it now, nothing should be broken by that.

koddsson:
Did you ever publish this @sejoker? I'm getting it in the diff when trying to publish now.

koddsson:
I just published this new variable as a patch version.

manatarms:
So the goal is to add a section to cf-styleguide (cf-designguide maybe?) that has a list of official colors, fonts and other design stuff. I'm working with @neynah on this right now.

Issue #204, opened 2017-05-24
Add gray theme for Pagination


koddsson:
This is the look of the backbone paginator currently running on cloudflare.com

![](https://cloud.githubusercontent.com/assets/318208/26365035/54a77258-3fde-11e7-9c68-5f4729b68a59.png)

In order for use to use this component in place of that backbone one, we need to style it the same.

koddsson:
This has been done for a while :D

PR #206, opened 2017-05-25
pagination: add a "minimal" gray theme

koddsson:
minimal-paginator

Related-to: #204


manatarms:
I just realized the hover cursor looked like a Mickey mouse glove.

koddsson:
This is really weird. I can't see the `PaginationLink` in any of those files :sob:

manatarms:
Do you think its importing something that is importing `PaginationLink`?

koddsson:
@manatarms I mean if you look at the jest ouput in Travis you see the files mentioned having this file but I can't see them in the files when I open 'em up locally. I'll take a better look to see if it is some weird importing thing.

manatarms:
Yeah the file that's failing in Jest requires `PaginationItem` which requires `PaginationLink`. But I cant find any other place that requires `PaginationLink` :(

koddsson:
I think I got this fixed!

The output from CI didn't help at all and apparently I didn't remove references to `PaginationLink`. I could have sworn I had tho 😱 

Hopefully this will fix the build and we can get this in.

koddsson:
Bam!

Any chance on getting this reviewed @manatarms @wyuenho @hturan @sejoker @tajo ?

packages/cf-component-pagination/src/PaginationItemMinimalTheme.js
```diff
@@ -0,0 +1,45 @@
...
+  link: {
+    userSelect: 'none',
+    position: 'relative',
+    display: 'block',
+    paddingTop: '.5em',
+    paddingBottom: '.5em',
+    paddingLeft: '.1em',
+    paddingRight: '.1em',
+    textDecoration: 'none',
+    fontWeight: 700,
```
manatarms: baseTheme maybe? Not sure.

koddsson: That's probably a good use case for the base theme.

koddsson:
Can I take the lack of feedback as approval? This is started to block me severely on my project.

@manatarms @wyuenho @hturan @sejoker @tajo

hturan:
I think my biggest concern here is that the gif above looks nothing like our current pagination:

screen shot 2017-05-31 at 11 34 49

I'd rather not have three distinct approaches in the dashboard if possible! 😊

packages/cf-component-box/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cf-component-box",
   "description": "Cloudflare Box Component",
-  "version": "1.4.3",
```
hturan: Why the MAJOR bump?

koddsson:
@hturan The font stuff?

packages/cf-component-box/package.json
koddsson: Hmmm. I didn't think I changed this as all 🤔

koddsson: I think Lerna should manage all this maybe? I'll reset these files.

packages/cf-util-http/src/http.js
```diff
@@ -184,10 +184,7 @@ export function request(method, url, opts, callback) {
 
         return response.text().then(text => {
           if (callback) {
-            callback(
-              undefined,
-              wrapResponse(headers, status, text, response)
-            );
+            callback(undefined, wrapResponse(headers, status, text, response));
```
hturan: This seems odd. Surely this line is too long for `prettier`?

packages/cf-component-box/package.json
hturan: Cool, thanks. There's a lot of unrelated package.json noise in here 👍

packages/cf-util-http/src/http.js
koddsson: Seems to fit. The horizontal line is the 80 char limit.

image

koddsson: 2 chars away from breaking that limit :D

hturan:
@koddsson Yeah, the type family, weight and spacing looks pretty off. Not sure if this is a global CSS issue, but I thought the point of `fela` was to avoid the global cascade?

hturan:
Looks good to me! If it works, it works. I definitely don't think I'm fully qualified though as I have no idea how themes and overrides work in `fela`, haha 😊

packages/cf-util-http/src/http.js
hturan: Haha, ace, my bad! Wonder why it was formatted differently before?

koddsson:
>@koddsson Yeah, the type family, weight and spacing looks pretty off. Not sure if this is a global CSS issue, but I thought the point of fela was to avoid the global cascade?

Yeah I've just been manually running these examples locally since I haven't checked out the internal styleguide to run this. I'll do that now and make sure the font's are correct.

packages/cf-component-box/package.json
```diff
@@ -1,7 +1,7 @@
...
+  "version": "2.0.0",
```
sejoker: somehow @koddsson you published version 2.0.0 of the package from this branch. I strongly disagree with publishing other than master branch. Our experience dealing with `legacy` branch showed that Lerna is not capable of resolving it properly.

packages/cf-component-box/package.json
koddsson: Yeah, not sure how that got in here 😨

packages/cf-component-box/package.json
koddsson: I published this?

sejoker: @koddsson you right, it looks suspicious cause publish appears in your branch commits. Please rebase from master to get latest packages versions.

sejoker:
@koddsson I see a lot of changes around PaginationLink refactoring to Link. along with some style changes in Link styles. Because cf-component-pagination is already in use in www-next we need to be extra careful not to break existing styles for `dot` mode. Could you explain why styles are changes for Link?

sejoker:
I feel like adding extra theme into package shouldn't be a breaking change. More reasoning on refactoring would be super helpful

packages/cf-component-box/package.json
koddsson: Yup. I was pulling from `origin`(my fork remote) and not `upstream`(this remote).

koddsson: Should be fixed now.

packages/cf-component-pagination/src/index.js
```diff
@@ -1,29 +1,40 @@
...
+const PaginationItemMinimal = applyTheme(
+  PaginationItemUnstyled,
+  PaginationItemMinimalTheme
+);
 
 export {
   Pagination,
-  PaginationItem,
-  PaginationRoot,
-  PaginationLink,
```
koddsson: @sejoker `PaginationLink` no longer exists === Breaking change?

koddsson:
I didn't notice when I took this screenshot but are the arrows misaligned?

packages/cf-util-http/src/http.js
koddsson: I took the `|| true` from the lint script and am running prettier on every file I change so it formatted this line correctly and am no longer allowing builds where prettier/eslint errors occur.

koddsson:
Build is failing because yarnpkg.com is down?

packages/cf-component-pagination/src/PaginationLink.js
```diff
@@ -1,74 +0,0 @@
...
-          onClick={this.handleClick}
-          aria-label={label}
-        >
-          {children}
-        </a>
-      );
-    }
-
-    return (
-      <span className={className} aria-label={label}>
```
sejoker: current implementation of Pagination link doesn't handle the case of rendering span than the link is not clickable.

packages/cf-component-pagination/src/PaginationLink.js
```diff
@@ -1,74 +0,0 @@
...
-
-  render() {
-    const { clickable, label, children, className } = this.props;
-
-    if (clickable) {
-      return (
-        <a
-          className={className}
-          href="#"
-          onClick={this.handleClick}
```
sejoker: implementation of pagination link doesn't call `e.preventDefault()` which causes the url change to `/#`

koddsson: I was thinking that instead of handling this for the user, they would take in the event in their click handler in case they want to do something with the events themselves.

packages/cf-component-pagination/src/PaginationLink.js
koddsson: What's "not clickable"? Is it that the cursor isn't a `pointer`?

sejoker: have you tested the changes on styleguide's `cf-component-pagination` example? I tried something similar on your branch and it didn't work for me either.

koddsson: Yeah I tested it and made a fix I just pushed. Couldn't verify the fix since the internal styleguide stopped working suddenly for me

koddsson:
@sejoker how about now?

sejoker:
@koddsson  feel free to merge when you are ready.

PR #207, opened 2017-05-25
Remove duplicate keys

koddsson:
This fix is brought to you by the @Microsoft typescript compiler:

```
packages/cf-component-heading/src/HeadingTheme.js(10,3): error TS1117: An object literal cannot have multiple properties with the same name in strict mode.
```

```
packages/cf-component-box/src/propertiesSpec.js(454,3): error TS1117: An object literal cannot have multiple properties with the same name in strict mode.
```

### DIFF
--- a/packages/cf-component-box/src/propertiesSpec.js
+++ b/packages/cf-component-box/src/propertiesSpec.js
@@ -449,11 +449,6 @@ export default {
       'unset'
     ]),
     default: 'normal'
-  },
-
-  wordWrap: {
-    propType: PropTypes.string,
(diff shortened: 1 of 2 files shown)


koddsson:
There seems to be another one of these in `propertiesSpec.js` but I'm not sure which one to remove:

image

koddsson:
Seems the former is more correct: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap?v=example

sejoker:
@koddsson thanks for the fix.

PR #232, opened 2017-05-26
introduce new gray color moonshine

sejoker:
This is a continuation on https://github.com/cloudflare/cf-ui/pull/203. I will deprecate now `colorGrayLightOnboarding` and remove it after migration to `moonshine`

### DIFF
--- a/packages/cf-style-const/src/variables.js
+++ b/packages/cf-style-const/src/variables.js
@@ -28,6 +28,7 @@ export default {
     dusk: '#4D4D4F',
     storm: '#808285',
     hail: '#BCBEC0',
+    moonshine: '#F7F7F7',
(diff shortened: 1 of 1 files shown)


koddsson:
I think my original question from #203 is applicable here too. Will the design team mark these colors on design for us so we know what color is `gray` and what color is `moonshine`?

sejoker:
@koddsson I use [InVision inspect](https://support.invisionapp.com/hc/en-us/articles/115000650823-Inspect) feature where you can explore all colors and ask to match the approved color set if colors don't match. So the answer is yes.

koddsson:
Yeah if getting the color names is trivial from given designs then I don't care what the name of the colors are :)

PR #233, opened 2017-05-30
Add support for 'id' prop on Card component

hturan:
In our current dashboard, lots of cards have an id prop which allows us to hash-scroll to them. This PR allows us to do the same thing with React `Card`s 🎉

Setting `id={undefined}` is fine, as React will just ignore and not render the `id` prop.

### DIFF
--- a/packages/cf-builder-card/test/__snapshots__/cardBuilder.js.snap
+++ b/packages/cf-builder-card/test/__snapshots__/cardBuilder.js.snap
@@ -3,6 +3,7 @@
 exports[`should render a card 1`] = `
 <section
   className="cf-card"
+  id={undefined}
 >
   <div
     className="cf-card__section cf-card__section--default"
@@ -32,6 +33,7 @@ exports[`should render a card 1`] = `
 exports[`should render a table 1`] = `
 <section
   className="cf-card"
+  id={undefined}
 >
   <div
     className="cf-card__section cf-card__section--default"
@@ -66,6 +68,7 @@ exports[`should render a table 1`] = `
 exports[`should render button 1`] = `
 <section
   className="cf-card"
+  id={undefined}
 >
   <div
     className="cf-card__section cf-card__section--default"
@@ -102,6 +105,7 @@ exports[`should render button 1`] = `
 exports[`should render drawers 1`] = `
 <section
   className="cf-card"
+  id={undefined}
 >
   <div
     className="cf-card__section cf-card__section--default"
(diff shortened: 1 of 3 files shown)


koddsson:
>Setting `id={undefined}` is fine, as React will just ignore and not render the `id` prop.

Weird that the snapshots have them in there then D:

hturan:
> Weird that the snapshots have them in there then D:

I don't understand? Snapshots render React, no?

hturan:
Anyone know why CI's failing?

koddsson:
I thought that snapshots where are a representation of the DOM when a component is mounted, but I guess that's not correct :)

hturan:
I've updated all snapshots by running `yarn test -- --u`, but it's failing to appease CI. Anyone know what's going on? Why does running `yarn test -- --u` not update `cf-builder-card` snapshots?

koddsson:
Could it be that `yarn test` only runs the changed files so it doesn't update all the snapshots? I always just run jest directly in watch mode and hit <kbd>a</kbd> to run all the tests before committing to make sure everything is updated.

Maybe we should update the script or the documentation?

hturan:
> I thought that snapshots where are a representation of the DOM when a component is mounted, but I guess that's not correct :)

Yeah, it outputs JSX, hence why you see `className` too 😊

> Could it be that yarn test only runs the changed files so it doesn't update all the snapshots? I always just run jest directly in watch mode and hit a to run all the tests before committing to make sure everything is updated.

It definitely runs everything. I get multiple failures locally, but that's tracked elsewhere.

koddsson:
Then I got no more suggestions of the top of my head 😢

hturan:
> Then I got no more suggestions of the top of my head 😢

I mean, the 'Merge' button is available. Are we happy to merge with red CI? I've got the correct # of approvals.

koddsson:
Do you mind if I check it out and try to muck about a bit? I just spent a lot of time getting some red tests green in a PR and I'd like to try to get PRs green if possible.

hturan:
Sure, feel free! Thanks!

koddsson:
Ah! We need to run build so that when we require a package it references the new code and not the last time code was built locally.

koddsson:
Hope it's cool I added a commit <3

koddsson:
🎉 🎉 🎉

hturan:
@koddsson yay cool thanks! I did run `yarn run bootstrap` and `yarn run build` before, but I guess it didn't catch. Oh well! I'll merge now 👍

PR #235, opened 2017-05-31
allow to render card link without drawer

sejoker:
screen shot 2017-09-21 at 4 13 45 pm

```
     <Card>
          <CardToolbar
            controls={[]}
            links={[
              <CardToolbarLink
                expandable={false}
                onClick={() =>
                  window.open('https://en.wikipedia.org/wiki/Main_Page')}
              >
                Wikipedia
              </CardToolbarLink>
            ]}
          />
    </Card>
```


koddsson:
Can we add a [link icon](https://www.iconfinder.com/icons/192090/external_go_to_hyperlink_link_new_tab_outside_url_icon#size=128) next to the link so it's apperent that it's a link that will open in the current or in a new page?

packages/cf-component-card/src/CardDrawerLink.js
```diff
@@ -0,0 +1,43 @@
...
+
+  handleClick(e) {
+    e.preventDefault();
+    this.props.onClick();
+  }
+
+  render() {
+    let className = 'cf-card__toolbar_link';
+
+    if (this.props.isActive) {
```
koddsson: Is `isActive` needed in a link where it doesn't open up a drawer?

packages/cf-component-card/src/CardToolbarLink.js
```diff
@@ -1,43 +1,33 @@
...
-        onClick={this.handleClick}
-      >
-        {this.props.children}
-      </Link>
-    );
-  }
-}
+const CardToolbarLink = ({ className, href, children, target }) => (
+  <a
+    className={`cf-link ${className}`}
```
koddsson: I think I'd like to have access to all the attributes of `<a />` or just pass in the `<a />` component myself.

Could we either spread `this.props` on this tag or change it to be just a decorative component and pass in the `<a />` ourselves?

packages/cf-component-card/src/CardToolbar.js
```diff
@@ -12,14 +12,18 @@ class CardToolbar extends React.Component {
         <div className="cf-card__toolbar_links" role="tablist">
           {this.props.links}
         </div>
+        <div className="cf-card__toolbar_links">
```
hturan: So any child passed to `CardToolbar` is now a link?

koddsson:
@sejoker Is there something we can do here to move this PR forward?

manatarms:
Any updates on what we're planning to do here.

packages/cf-component-card/src/CardToolbarLink.js
```diff
@@ -36,25 +38,33 @@ class CardToolbarLink extends React.Component {
...
   render() {
+    const { id, className, to, children } = this.props;
     return (
       <Link
         role="tab"
-        id={this.props.id}
-        className={this.props.className}
+        id={id}
+        className={className}
+        to={to}
```
marksteyn: The role should be "link" if opens a link rather than a drawer

packages/cf-component-card/src/CardToolbarLink.js
```diff
@@ -3,25 +3,27 @@ import PropTypes from 'prop-types';
...
-    fontWeight: 'normal',
-    marginTop: '-0.7em',
-
-    position: 'absolute',
-    top: '51%',
-    right: '1rem',
-    transition: 'transform 0.2s ease',
-    transform: isActive ? 'rotate(90deg)' : 'initial'
-  }
+  '&::after': expandable
```
hturan: Could we use the `Icon` component here?

sejoker: It's not so obvious, (I would say it's impossible) because of css animation we use to expande/collapse motion.

packages/cf-component-card/example/basic/component.js
```diff
@@ -58,6 +60,20 @@ class CardComponent extends React.Component {
             }
           ]}
         />
+        <CardSection>
```
hturan: Is the `<CardSection>` wrapper necessary here?

packages/cf-component-card/src/CardToolbarLink.js
manatarms: I'm going make this a common component soon. I'll swap out the places that use it. We have rotating link one offs everywhere.

hturan: @manatarms Yep, this is such a common pattern now. That would be ace!

packages/cf-component-card/src/CardToolbarLink.js
sejoker: fixed, thank you.

packages/cf-component-card/example/basic/component.js
sejoker: apparently it's not.

PR #238, opened 2017-06-02
Add z-index max and modal boxshadow

prestonp:


packages/cf-style-const/src/variables.js
```diff
@@ -95,5 +95,7 @@ export default {
   colorPurpleDark: darken(colorOffsetDark, '#d91698'),
   colorTwitterBlue: '#00aced',
   colorFacebookBlue: '#3b5998',
-  colorMainBackground: '#e6e6e6'
+  colorMainBackground: '#e6e6e6',
+  zIndexMax: 1000,
+  modalBoxShadow: '0 0 0 9999px rgba(0, 0, 0, .7)'
```
manatarms: Is there a reason this is different from our current modal box shadow of `0 1px 15px rgba(0,0,0,.75)`?

tajo:
You should fix `author` field of your commit. :)

packages/cf-style-const/src/variables.js
prestonp: That box-shadow looks like a soft drop shadow, where as the appstore modal is using this giant shadow to implement a page fade.

screen shot 2017-06-02 at 5 01 00 pm

prestonp:
@tajo D'oh my cf email wasn't linked to my gh account. Fixed that up, thanks!

PR #240, opened 2017-06-05
Use em for 'dots' pagination

sejoker:
dots looks weird with fontSize 15px, font size should be an even number.

with fontSize 15px:
screen shot 2017-06-05 at 2 51 52 pm

with fontSize 16px:
screen shot 2017-06-05 at 2 52 24 pm

After this change we could set fontSize to 16px on the parent container.


sejoker:
@koddsson with current styles we won't see a difference. I am planning to use 16px for my private component which uses pagination component.

PR #244, opened 2017-06-06
fix: make the margins between pagination items larger

koddsson:
Make em bigger.

I took the value from our existing codebase and it looks like this:

image


koddsson:
err.. did you also publish this @marksteyn ?

PR #245, opened 2017-06-06
Update cf-util-http so the promises returned resolve to json

marksteyn:
The fetch promise resolves to a stream with methods such as text() and json(). These methods cannot be called twice, and text() is called internally in the request method. This means that response body can only be accessed in the callback and not via the object resolved by the promise.

This PR updates the request method so that the promise resolves with the same object that's passed to the success callback, and throws with the same object that's passed to the error callback.

### DIFF
--- a/packages/cf-util-http/src/http.js
+++ b/packages/cf-util-http/src/http.js
@@ -68,13 +68,7 @@ function wrapResponse(headers, status, text, response) {
   if (isJSON(headers['content-type'])) {
     body = text ? JSON.parse(text) : undefined;
   }
-  return {
-    headers,
-    status,
-    body,
-    text,
-    response
-  };
+  return { headers, status, body, text, response };
 }
 
 const BODYLESS_METHODS = ['GET', 'HEAD'];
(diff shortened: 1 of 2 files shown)


packages/cf-util-http/src/http.js
```diff
@@ -114,12 +108,10 @@ const BODYLESS_METHODS = ['GET', 'HEAD'];
  * @see [Global Fetch on MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch)
  * @see [WHATWG Fetch API Spec](https://fetch.spec.whatwg.org/)
  */
-export function request(method, url, opts, callback) {
+export function request(method, url, opts = {}, callback) {
   if (typeof opts === 'function') {
     callback = opts;
     opts = {};
-  } else if (opts == null) {
```
wyuenho: Keep this, otherwise passing a `null` to opts will set opts to null instead of `{}`

packages/cf-util-http/test/http.js
```diff
@@ -63,6 +63,18 @@ describe('request', () => {
     });
   });
 
+  test('should accept 3-args form where the opts is optional (promise)', done => {
```
wyuenho: 2-args

PR #246, opened 2017-06-06
Added package-lock.json to gitignore

manatarms:
TLDR; This PR ignores the auto generated NPM5 `package-lock.json` file.

Not that NPM@5 is out, it generates a `package-lock.json` file by default. But since we're using yarn we don't need the `package-lock.json`. 

If anyone feels strongly about committing the `package-lock.json` to the repo, I'll open another PR that commits the `package-lock.json` files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ packages/*/npm-debug.log.*
 es
 dist
 \.jest
+package-lock.json


jwineman:
How strict do we want to be on this? We don't use NPM on CF-UI and if we did we aren't using NPM5.  

Obviously there isn't any harm in merging this but I sort of lean toward declining it since we want people using Yarn.

manatarms:
@jwineman My bad. My wording was confusing. This PR ignores the `package-lock.json` file generated by NPM5.

jwineman:
I know.  My point is we want people using yarn, not NPM 5, adding the lock file to gitignore sends a mixed message.

manatarms:
Ohhh. I see what you mean. The annoy thing is that NPM generates the files whenever it touches node_modules or package.json. I was doing some npm stuff (npm version patch I think) and it decided to make a lock file for all of our packages.

PR #247, opened 2017-06-06
Update modal backdrop

prestonp:


packages/cf-style-const/src/variables.js
```diff
@@ -97,5 +97,5 @@ export default {
   colorFacebookBlue: '#3b5998',
   colorMainBackground: '#e6e6e6',
   zIndexMax: 1000,
-  modalBoxShadow: '0 0 0 9999px rgba(0, 0, 0, .7)'
+  modalBackdropBackground: 'rgba(0, 0, 0, .7)'
```
prestonp: Hey no worries! This is for a backdrop that covers the entire page. So it's a completely different pattern

Applied to a gh comment
```css
box-shadow: 0 1px 15px rgba(0,0,0,.75);
```

Cover entire page
screen shot 2017-06-06 at 2 44 28 pm

jwineman: We probably need `rgba()` in a color const in `cf-style-const` too right?

adamschwartz: You guys use `background: #000; opacity: .7` for your modal backdrop. So perhaps you’ll want to copy those two constants from there.

manatarms: Hmm. I think there might be a difference between our react and bb modals.

prestonp: @jwineman Not sure I understand, you want `rgba(0,0,0,0.7)` as a generic color const that's unrelated to modals?

Maybe something like overlayColor: 'rgba(0,0,0,0.7)` ?

adamschwartz: > Maybe something like overlayColor: 'rgba(0,0,0,0.7)' ?

That would be very reasonable to me. If you want this to be in sync with the existing modals in the CF Dash (not sure whether those are "react" or "bb" with @manatarms  terminology) though, you’d want to find out where the existing `background: #000` and `opacity: .7` are coming from.

jwineman: As a general rule we try to keep colors in variables so we can track all the different colors we use across the app and eventually standardize them.  I think we should put this `rgba` in [`variables.js`](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-const/src/variables.js)

prestonp:
@jwineman Renamed the rgba color in variables.js

packages/cf-style-const/src/variables.js
adamschwartz: That makes sense to me. Then my recommendation would just be the inverse (that those other modal styles should instead use this new variable). Cuz, DRY.

jwineman:
I'm an idiot, I thought you were editing modal, not this package.  My bad.

PR #251, opened 2017-06-12
Fix for border radius for spaced buttons

manatarms:
This fixes an edge case where the border radius of spaced button is not maintained.


koddsson:
@manatarms Has this been published?

manatarms:
Yup @koddsson

koddsson:
Nice, I figured as much but I like to remind people just in case since we keep forgetting 😛

PR #252, opened 2017-06-12
Added height to theme to override

manatarms:
Need this height to override the theme in a few places.


tajo:
@manatarms Why do we need a variable height for the Button?

manatarms:
After talking with @tajo I realized I don't need height anymore. Thanks @tajo

PR #254, opened 2017-06-13
Added button loading animation

manatarms:
Added a loading animation to the button.


packages/cf-component-button/src/Button.js
```diff
@@ -59,19 +65,24 @@ const marginLeft = (marginLeft, group, spaced) => {
 
 const styles = props => {
   const theme = props.theme;
+  console;
```
sejoker: please remove

sejoker:
adding a gif video with end result would be nice.

manatarms:
@sejoker Design doesn't want to use the loading gif. They approved this animation instead.

Issue #219, opened 2017-05-26
Migrate `cf-component-table` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

toekneestuck:
Moved internal.

PR #256, opened 2017-06-14
#219 migrate cf-component-table to fela

wyuenho:
This PR migrates all of our existing SASS and PostCSS table styles to cf-component-table via Fela.

This migration slightly deviates from the pattern of currently migrated components in the following ways:

1. Despite the name, all of the those "Unstyled" components are actually unthemed, not unstyled. ~~This PR completely separates the styles from the components - component theme for minor customization, and style functions for replacing the styling completely.~~
~~2. Due to 1, the `className` prop remains open, in addition to `styles`. You should not abuse `className`. The relationship between them is either/or. You can either use a `className` automatically applied by `createComponent`, or the className that comes from the `styles` object given by `createComponentStyles`, but not both at the same time.~~
3. To achieve high fidelity to our existing selector-heavy styles, a number of creative ways to "emulate" CSS selectors had to be invented. The indices of allowed ancestors are passed all the way down to the cells, so we could simulate `:first-child`, `:last-child`, and `:nth-child` and so on. `:last-child` is signal by virtue of being an index that has a negative number. These indices are passed down by cloning the children and adding these "CSS query" props at every level. A number of helper functions were introduced into `cf-style-container` to help with is.
4. Separately, the inner logic of the `applyTheme` HOC had been extracted to ease testing. In addition, `applyTheme` can now take object themes instead of just function themes. This brings our wrapper layer back to feature parity with vanilla Fela's `<ThemeProvider>`. 
~~5. Arbitrary props value for the underlying DOM elements remain open, with the exception of the 
`style` attribute.~~

There are mainly 3 reasons to completely separate the styles from the structure of the components.
1. There will be new internal components introduced that rely on table structures while styled completely differently. Separating the style functions from the components allows this possibility.
2. Design is currently experimenting with a new design, so it's conceivable that we might have 2 separate styles living side-by-side. Decoupling the styles from the structure allows maximum room for progress on the new design.
3. It's often needed to use `createComponent`, `createComponentStyles` and `applyTheme` to create one off override to some style properties of presentational components.

I hope to promote this pattern in general because this pattern can drastically reduce stray `<div>`s generated by `<Box>`. Imagine instead of having to wrap every block you need to layout with a `<Box>`, you could import the style functions and themes separately, and compose in your overrides, and then apply to the blocks at issue.

# Note:
~~Since this package now requires direct parent/child relationship and runtime type detection among components to work, if you do need to mix in you custom table components with any components provided by this package, please make sure you ***inherit*** the one you intend to customize.~~

(I'll drop in some usage examples to explain this more tomorrow)

### DIFF
--- a/packages/cf-component-table/package.json
+++ b/packages/cf-component-table/package.json
@@ -11,9 +11,13 @@
     "registry": "http://registry.npmjs.org/"
   },
   "dependencies": {
+    "cf-style-container": "^1.3.1",
     "prop-types": "^15.5.8"
   },
   "peerDependencies": {
     "react": "^0.14.9 || ^15.0.0-0 || ^16.0.0-0"
+  },
+  "devDependencies": {
+    "cf-style-provider": "^2.0.0"
   }
 }
--- a/packages/cf-component-table/src/Table.js
+++ b/packages/cf-component-table/src/Table.js
@@ -1,41 +1,90 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import {
+  mapChildren,
+  filterNone,
+  applyTheme,
+  createComponent,
+  connectStyles
+} from 'cf-style-container';
+import TableTheme from './TableTheme.js';
+
+const TableStyles = ({ theme, bordered }) =>
+  filterNone({
+    background: theme.colorWhite,
+    borderCollapse: bordered ? 'separate' : 'collapse',
+    borderSpacing: 0,
+    maxWidth: '100%',
+    width: '100%',
+
+    // Table styles
+    border: bordered ? '1px solid ${theme.tableBorderColor}' : undefined,
+    borderLeft: bordered ? 0 : undefined,
+
+    [`@media (max-width: ${theme.breakpoints.tablet})`]: {
+      display: 'block',
+      width: '100%',
+      boxSizing: 'border-box'
+    },
+
+    [`@media screen and (max-width: ${theme.breakpoints.tablet})`]: {
+      display: 'block'
+    }
+  });
 
 class Table extends React.Component {
   render() {
-    let className = 'cf-table';
+    const {
+      is,
+      className,
+      striped,
+      hover,
+      bordered,
+      condensed,
+      bare,
+      children
+    } = this.props;
 
-    if (this.props.striped) className += ' cf-table--striped';
-    if (this.props.hover) className += ' cf-table--hover';
-    if (this.props.bordered) className += ' cf-table--bordered';
-    if (this.props.condensed) className += ' cf-table--condensed';
-    if (this.props.className.trim())
-      className += ' ' + this.props.className.trim();
-
-    return (
-      <table className={className}>
-        {this.props.children}
-      </table>
+    return React.createElement(
+      is,
+      { className },
+      mapChildren(children, child => {
+        if (React.isValidElement(child)) {
+          return React.cloneElement(child, {
+            condensed,
+            striped,
+            bordered,
+            hover,
+            bare
+          });
+        }
+        return child;
+      })
     );
   }
 }
 
 Table.propTypes = {
-  className: PropTypes.string,
+  is: PropTypes.oneOf(['table', 'div']),
+  className: PropTypes.string.isRequired,
   striped: PropTypes.bool,
   hover: PropTypes.bool,
   bordered: PropTypes.bool,
   condensed: PropTypes.bool,
+  bare: PropTypes.bool,
   children: PropTypes.node
 };
 
 Table.defaultProps = {
-  className: '',
+  is: 'table',
   striped: false,
   hover: false,
   bordered: true,
-  condensed: false
+  condensed: false,
+  bare: false
 };
 
-export default Table;
+export const createTable = (...styles) =>
+  applyTheme(connectStyles(TableStyles, ...styles)(Table), TableTheme);
+
+export default createTable();
--- a/packages/cf-component-table/src/TableCell.js
+++ b/packages/cf-component-table/src/TableCell.js
@@ -119,26 +119,7 @@ export const TableCellStyles = ({
 
 class TableCell extends React.Component {
   render() {
-    const {
-      is,
-      className,
-      children,
-      condensed,
-      striped,
-      bordered,
-      hover,
-      bare,
-      rowType,
-      theadIndex,
-      tbodyIndex,
-      tfootIndex,
-      rowIndex,
-      cellIndex,
-      align,
-      width,
-      rowSpan,
-      colSpan
-    } = this.props;
+    const { is, className, children, rowSpan, colSpan } = this.props;
 
     return React.createElement(
       is,
@@ -147,7 +128,7 @@ class TableCell extends React.Component {
         rowSpan: is === 'div' ? undefined : rowSpan,
         colSpan: is === 'div' ? undefined : colSpan
       },
-      this.props.children
+      children
     );
   }
 }
--- a/packages/cf-component-table/src/TableCell.js
+++ b/packages/cf-component-table/src/TableCell.js
@@ -1,37 +1,186 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import {
+  filterStyle,
+  capitalize,
+  filterNone,
+  applyTheme,
+  createComponent,
+  connectStyles
+} from 'cf-style-container';
+import TableTheme from './TableTheme';
+
+export const TableCellStyles = ({
+  is,
+  theme,
+  condensed,
+  striped,
+  bordered,
+  hover,
+  bare,
+  rowType,
+  theadIndex,
+  tbodyIndex,
+  tfootIndex,
+  rowIndex,
+  cellIndex,
+  align,
+  width
+}) =>
+  filterNone({
+    display: is === 'div' ? 'inline-block' : undefined,
+    flex: is === 'div' ? 'auto' : undefined,
+
+    // Baseline cell styles
+    borderTop: ((theadIndex === 0 || tbodyIndex === 0) && rowIndex === 0) ||
+      bare
+      ? 0
+      : `1px solid ${theme.tableBorderColor}`,
+    lineHeight: theme.lineHeight,
+    padding: condensed ? theme.tableCondensedPadding : theme.tablePadding,
+    verticalAlign: 'middle',
+    textAlign: align,
+    whiteSpace: 'nowrap',
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+
+    width: width,
+
+    // For first th/td in the first row in the first thead or tbody
+    borderTopLeftRadius: tbodyIndex === 0 && rowIndex === 0 && cellIndex === 0
+      ? theme.borderRadius
+      : undefined,
+    // The last index is passed down as its complement so you can detect whether
+    // it's the last child easily while preserving its magnitude
+
+    // For last th/td in the first row in the first thead or tbody
+    borderTopRightRadius: tbodyIndex === 0 && rowIndex === 0 && cellIndex < 0
+      ? theme.borderRadius
+      : undefined,
+
+    // For first th/td (can be either) in the last row in the last thead, tbody, and tfoot
+    borderBottomLeftRadius: (tbodyIndex < 0 || tfootIndex < 0) &&
+      rowIndex < 0 &&
+      cellIndex === 0
+      ? theme.borderRadius
+      : undefined,
+
+    // For last th/td (can be either) in the last row in the last thead, tbody, and tfoot
+    borderBottomRightRadius: (tbodyIndex < 0 || tfootIndex < 0) &&
+      rowIndex < 0 &&
+      cellIndex < 0
+      ? theme.borderRadius
+      : undefined,
+
+    // Clear border-radius for first and last td in the last row in the last
+    // tbody for table with tfoot
+    borderRadius: tfootIndex != null &&
+      tbodyIndex < 0 &&
+      rowIndex < 0 &&
+      (cellIndex === 0 || cellIndex < 0)
+      ? 0
+      : undefined,
+
+    // Responsive styles
+    [`@media (max-width: ${theme.breakpoints.tablet})`]: {
+      display: 'block',
+      width: '100%',
+      boxSizing: 'border-box',
+      borderTopWidth: 0,
+      borderTop: cellIndex === 0
+        ? `1px solid ${theme.tableBordercolor}`
+        : undefined
+    },
+
+    // Row Styles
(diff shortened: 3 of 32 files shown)


.babelrc
```diff
@@ -5,6 +5,7 @@
   ],
   "plugins": [
     "transform-object-rest-spread",
+    "transform-runtime",
```
koddsson: What's this?

wyuenho: This is basically babel-polyfill for libraries. babel parses your code and look for any ES6 method calls, and inline those calls into the compiled bundle. This also removes all those _extend calls in the compiled code as well.

The only reason this is here is because I'm using Object.entries() for convenience.

packages/cf-component-table/src/TableBodyStyle.js
```diff
@@ -0,0 +1,15 @@
+export default ({ theme }) => ({
+  // TODO
```
jwineman: Is this still a TODO?

jwineman:
> Due to 1, the className prop remains open, in addition to styles. You should not abuse className. The relationship between them is either/or. You can either use a className automatically applied by createComponent, or the className that comes from the styles object given by createComponentStyles, but not both at the same time.

This is not a slight deviation.  Can you go into more detail around your first point?  The exported modules for other fela migrated components are styled with the default theme, I don't understand why table is any different. 

One large reason we chose fela is because it allows us to **white list the css properties that are customizable so we can avoid exposing `className`.**

packages/cf-component-form/test/__snapshots__/FormFieldset.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`should render 1`] = `
 <fieldset
-  className="cf-cy1e5a"
```
wyuenho: I don't understand why my local build fails without this, while CI says otherwise.

tajo:
First of all, thanks for the work. This is a big and important component.

> This migration slightly deviates from the pattern of currently migrated in the following ways

It deviates a lot because

> the className prop remains open

and that goes against one of the [founding principles of cf-ui](https://github.com/cloudflare/cf-ui/blob/b748590e7b625ce9cdec1ec90a942492c7682a69/discussions/no-customizing-classnames.md).

For this reason alone, I can't support this PR and **I would really like to see a technique oriented DIP that will present specific examples**. We need to stay consistent across the whole library and this is a complete one-off compared to other migrated components that follow exactly same style and structure.

> There are mainly 3 reasons to completely separate the styles from the structure of the components.

Those reasons are not compelling to me since they are speculative at this point:

> we might have 2 separate styles living side-by-side

Also, not very specific.

> There will be new internal components introduced that rely on table structures while styled completely differently.

Again, it's not a current problem. We might come to conclusion that is better to create new components instead rather than trying to fit "completely differently styled" table like components into cf-component-table that is already quite complex.

tajo:
# How to customize our components

1. **Add a new prop**. This is something we should always strive for since it creates clear and transparent interface. If there are too many props already, the component probably does too many things and it should be split / reevaluated.

2. **Overwrite the component theme**. This is a new option with fela. Migrated components create CSS rules based on props and **theme values**. We always provide a default theme and export our components tied to it. However, you can do

```js
const MyButton = applyTheme(ButtonUnstyled, ButtonTheme, () => { color: "red" });
<MyButton />
```

and change the color for cf-component-button. What values can you customize this way? Only values that are abstracted into `ButtonTheme`. If `color` wasn't there and `Button` component defined the style function as

```js
rules  = (props) => ({
  color: 'blue'
})
```

instead of 

```js
rules  = (props) => ({
  color: props.theme.color
})
```

You would not be able to customize it. **Theme gives us a tool to whitelist values that can be customizable**. Still, new props are almost always better solution but sometimes we need to do one-offs and this is the way how to do it. Also, note that theme is a set of constants passed through React's context. It can't be ever updated. Props need to used instead.

^^ **Both those ways use whitelisting approach.** It needs to be allowed by our code (via props or theme). Note: Themes now abstract almost all values for CSS properties that are used by default. So you would need to create a PR only if you want to customize through a CSS rule that's not even used currently.

---

3. **Open className prop** / **Make the style function swappable** - This way you can overwrite absolutely everything and even add more styles that were not defined before. **There is zero control and every instance of cf-ui component can go completely wild, no restrictions for developers**. This approach [was not allowed](https://github.com/cloudflare/cf-ui/blob/b748590e7b625ce9cdec1ec90a942492c7682a69/discussions/no-customizing-classnames.md) even before the migration. Considering themes, it's even less reasonable now.

packages/cf-component-table/src/Table.js
```diff
@@ -1,41 +1,102 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { filterStyle, filterNone, mapChildren } from 'cf-style-container';
+import TableHead from './TableHead';
+import TableBody from './TableBody';
+import TableFoot from './TableFoot';
+
+function issubclass(childType, Type) {
```
marksteyn: isSubclass for consistency

packages/cf-style-const/src/variables.js
```diff
@@ -4,6 +4,13 @@ const colorOffsetLight = 0.15;
 const colorOffsetDark = 0.07;
 
 export default {
+  breakpoints: {
+    mobile: '@media (min-width: 13.6em)',
+    mobileWide: '@media (min-width: 30.4em)',
```
sejoker: it's unmaintainable to support 5 media breakpoints, in my experience it's a lot of effort to work with 2 points. Where this list come from? We should have a conversation with design team and agreed what we support.

packages/cf-style-const/src/variables.js
```diff
@@ -4,6 +4,13 @@ const colorOffsetLight = 0.15;
 const colorOffsetDark = 0.07;
 
 export default {
+  breakpoints: {
+    mobile: '@media (min-width: 13.6em)',
```
sejoker: let's fix values in px to have a consistent behavior.

packages/cf-component-table/src/Table.js
```diff
@@ -1,41 +1,102 @@
...
+
+            return React.cloneElement(
+              child,
+              filterNone({
+                condensed,
+                striped,
+                bordered,
+                hover,
+                bare,
+                theadIndex: issubclass(child.type, TableHead)
```
sejoker: could we have tests for theadIndex, tbodyIndex, tfootIndex counters logic?

sejoker:
I don't buy the idea of exposing styles property for the future needs. As @tajo pointed out we should go through specific examples to understand the limitation of the current approach with themes and Fela's generated classname. I would aim for consistency in a way we use Fela, we already have a set of practices which worked pretty well so far. For you idea with styles we probably need a DIP.

packages/cf-style-const/src/variables.js
wyuenho: They come from our existing SASS and PostCSS variables files. I don't like it, but I also don't want to change it for fear of making the migration even harder.

packages/cf-style-const/src/variables.js
wyuenho: Let's deal with this later. This is a pretty large and overarching change, I want to understand why they are in ems in the first place.

packages/cf-component-table/src/Table.js
wyuenho: What do you mean?

packages/cf-style-const/src/variables.js
tajo: Apparently ems are way to go. @mrmrs can explain.

packages/cf-style-container/src/index.js
```diff
@@ -18,6 +19,16 @@ const createComponent = (rule, type = 'div', passThroughProps = []) =>
       : passThroughProps
   );
 
+const connectStyles = (...styles) => type =>
```
sejoker: I see, instead of spreading themes you spread styles, neat technique.

packages/cf-component-table/src/TableCell.js
```diff
@@ -1,37 +1,186 @@
...
 
-    return (
-      <td className={_className} {...props}>
-        {this.props.children}
-      </td>
+class TableCell extends React.Component {
+  render() {
+    const {
+      is,
+      className,
```
sejoker: why do you need to destructure all props here?

packages/cf-component-table/src/TableCell.js
```diff
@@ -1,37 +1,186 @@
...
 
 TableCell.defaultProps = {
-  className: ''
+  is: 'td',
+  align: 'left'
 };
 
-export default TableCell;
+export const createTableCell = (...styles) =>
+  applyTheme(connectStyles(TableCellStyles, ...styles)(TableCell), TableTheme);
```
sejoker: the limitation I see with using `connectStyles`: styles won't get access to props, e.g. `theadIndex`,`tbodyIndex` etc.

wyuenho: You do get access to it, we are using our wrapped `createComponent`, which will automatically passThru props if defined in propTypes.

packages/cf-component-table/src/TableCell.js
wyuenho: left-over code from where I was spreading props

packages/cf-component-table/package.json
```diff
@@ -11,9 +11,13 @@
...
   },
   "dependencies": {
+    "cf-style-container": "^1.3.1",
     "prop-types": "^15.5.8"
   },
   "peerDependencies": {
     "react": "^0.14.9 || ^15.0.0-0 || ^16.0.0-0"
+  },
+  "devDependencies": {
+    "cf-style-provider": "^2.0.0"
```
tajo: There's no reason for devDeps since deving/testing is managed from the root level of cf-ui. Also, less dependencies = less version bumps by lerna.

packages/cf-component-table/src/Table.js
```diff
@@ -1,41 +1,90 @@
...
   striped: false,
   hover: false,
   bordered: true,
-  condensed: false
+  condensed: false,
+  bare: false
 };
 
-export default Table;
+export const createTable = (...styles) =>
```
tajo: This should not be exported because it can be used as

```js
import { createTable } from 'cf-component-table';

const Table = createTable(() => { absolutely_any_styles_I_want_to_use_I_can_put_here });
```

So that's exactly same as allowing appending a custom className in CSS world.

On the other hand, TableTheme is directly imported by the component and not exposed to for overwriting at all.

So more restricted and controlled access to customization (theme) is replaced by "do whatever you want to do" `styles` param. **It should be the other way around.**

My believe is that developers should not have that level of access because then the "owners" of cf-ui are completely losing control over the styles and it would be impossible for them to refactor/improve without introducing breakages.

packages/cf-style-container/src/index.js
```diff
@@ -8,6 +8,7 @@ import {
 } from 'react-fela';
 import { static as Immutable } from 'seamless-immutable';
 import { capitalize } from 'underscore.string';
+import mergeOptions from 'merge-options';
```
wyuenho: Argh, looks like a bad merge

sejoker: I was surprised to see it actually.

packages/cf-component-table/src/Table.js
sejoker: I disagree that we should limit styles extensibility only to `theming`. It should be *the primary* way but not the only one. I think we will hit nice balance if we leave both options available to the developers. It sounds a bit exaggerated:
>  it would be impossible for them to refactor/improve without introducing breakages

More importantly as Fela migration demonstrated major refactoring will produce breakage with a high probability anyway. `cf-ui` should make default usage simple and advanced case possible.

wyuenho: I haven't had the time to go thru this in detail, so I'll just briefly summary what's in my head now.

First, overriding styles externally is an inevitability, so it's a necessity, I don't think we disagree on this.

The disagreement mainly is the degree we allow customization. You seem to be misled that whitelisting style props in a theme makes it harder to override a style prop that's not allowed in a theme. Try this:

```jsx
import { Button } from 'cf-component-button';
const MyButton = createComponent(() => ({background: 'black'}), Button);
``

This button is themed, and `background` [is not whitelisted](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-component-button/src/Button.js). What actually gets generated is 2 fela generate CSS rule names applied to the button like this:

```html
<style>
.cf-14a9t3k {
  /* the original backgroundColor prop got DCE'ed and disappeared */
}
.cf-1e4pz7w {
  background: 'color';
}
</style>
<button type="button" class="cf-14a9t3k cf-1e4pz7w">
```

Even if fela didn't do this, overriding the button's style is not hard at all. I can do a number of things similar to this:

```jsx
class MyButton extends React.Component {
    componentWillMount() {
      const style = document.createElement('style');
      style.innerText = ".my-button { background-color; 'black' !important }";
      document.body.insertBefore(style, document.body.firstChild);
    }
    render() {
        return (
          <div className="my-button">
            <Button>
          </div>
        );
    }
}
```

There are 2 points I want to get across.

1. There's no way you can prevent me from overriding any style I want, when I want, and you are not making it any harder either.
2. Themes, are used as place holders for values derived from the base theme, and nothing more. We don't ever override these theme values in our existing SASS and PostCSS stylesheets.

There's more I have to say about themes in our ported components now, like the button. We are essentially abusing it as a style function. I'll save this for another day as this reply is already too long.

jwineman: > First, overriding styles externally is an inevitability, so it's a necessity, I don't think we disagree on this.

Agreed, where we disagree is whether or not the **overriding of styles should happen through fela themes**.

>    There's no way you can prevent me from overriding any style I want, when I want, and you are not making it any harder either.

The goal isn't to make it impossible, we want to make it hard to make bad choices and easy to make good ones.  By not exposing `style` and `className` externally it is easier to extend the fela theme rather than inject the styles "incorrectly".

wyuenho: Again, it's not hard to override the styles at all, as illustrated in my last comment.

Read the PR again. `className` must be exposed for Fela to apply the theme function.

Let me know if you can find a way to lock down styles, or at least make it a little harder so I can adjust this PR to achieve what you want.

jwineman: This is a trade off of using Fela, an architectural decision we've _already made_.  Its important we use the framework the way it was intended.  If you disagree you can submit a D.I.P.   This isn't something I consider up for discussion otherwise.

tajo: @wyuenho https://github.com/cloudflare/cf-ui/pull/278 fixes "passing" className through nested createComponent

PR #262, opened 2017-06-21
Styling utils improvements

wyuenho:
This PR is extracted from #256 

Improvements:
- breakpoint constants
- applyTheme now takes an object as the theme as well as a function
- A number of convenient functions to batch process style objects


packages/cf-style-container/package.json
```diff
@@ -12,9 +12,10 @@
   },
   "dependencies": {
     "fela": "^5.0.0",
-    "merge-options": "0.0.64",
+    "merge-options": "^1.0.0",
+    "prop-types": "^15.5.8",
     "react-fela": "^5.0.0",
-    "prop-types": "^15.5.8"
+    "underscore.string": "^3.3.4"
```
marksteyn: What's capitalize being used for? It seems that the only reason for including underscore.string is to export it. Why not just import it where you need it?

packages/cf-style-container/src/index.js
```diff
@@ -49,9 +56,40 @@ const createComponentStyles = (styleFunctions, component) =>
...
+
+// Filter out all the style attribute from the props
+export const filterStyle = props =>
+  filterProps(props, (key, value, accum) => {
+    if (key === 'style') return accum;
+    accum[key] = value;
+    return accum;
+  });
+
+export const mapChildren = (children, func) =>
```
marksteyn: What about renaming this mapChildrenWithIndex to indicate why it differs from React.Children.map?

sejoker:
@wyuenho could you rebase from master and update test snapshots? Looks like I broke tests with latest merge

packages/cf-style-container/src/index.js
wyuenho: You also get a real array tho. It's too cumbersome to put types in names.

marksteyn: The point was to differentiate from React.Children.map

packages/cf-style-container/package.json
wyuenho: This is often needed in construct the names of some of our existing CSS rules like `tableStripedDefaultBackground`. Since underscore.string comes with tons of functions, it's too easy to import the entire library and prevent webpack from treeshaking well. It's better to not tell people we are using underscore.string. If you need to capitalize, just use the one in cf-style-container. This makes it easier to swap out underscore.string too.

packages/cf-style-container/src/index.js
wyuenho: Hence `mapChildren`.

PR #264, opened 2017-06-21
Button float

manatarms:


packages/cf-style-container/test/__snapshots__/index.js.snap
```diff
@@ -27,10 +27,6 @@ Object {
     "boxShadow": "0 0 20px 0 rgba(136,136,136,0.50)",
```
manatarms: Not sure why this snapshot update but I guess the snapshot wasn't updated. Any ideas?

PR #268, opened 2017-06-22
Migrate Page to fela and add PageContent component

jculvey:
This does a few things:
  1. Migrates Page to fela
  2. Adds a new PageContent component for handling horizontal spacing

### DIFF
--- a/packages/cf-component-page/src/PageHeader.js
+++ b/packages/cf-component-page/src/PageHeader.js
@@ -1,22 +1,41 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
 
-class PageHeader extends React.Component {
-  render() {
-    return (
-      <header className="cf-page__header">
-        <h1 className="cf-page__title">{this.props.title}</h1>
-        {this.props.subtitle &&
-          <h2 className="cf-page__subtitle">{this.props.subtitle}</h2>}
-      </header>
-    );
-  }
-}
+const PageHeading = createComponent(
+  () => ({
+    fontSize: '2rem',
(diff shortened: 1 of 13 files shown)


packages/cf-component-page/src/PageHeader.js
```diff
@@ -1,22 +1,38 @@
...
+const PageSubheading = createComponent(
+  ({ theme }) => ({
+    ...theme.subheading
+  }),
+  'h2'
+);
+
+const PageHeader = ({ title, subtitle, className }) => {
+  return (
+    <header className={className}>
```
koddsson: Why a `className` here and not use `fela`?

tajo: `PageHeader` is wrapped by `createComponent` before exporting, so it uses fela and doesn't expose className.

packages/cf-component-page/src/PageHeader.js
```diff
@@ -1,22 +1,41 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const PageHeading = createComponent(
+  () => ({
+    fontSize: '2rem',
+    lineHeight: '1.2',
+    fontWeight: 400
```
manatarms: style-const?

packages/cf-component-page/src/PageHeader.js
koddsson: Is the `className` here provided by `fela`? This is the same thing as doing:

```js
createComponent(
  () = > {},
  'header',
  ['title', 'subtitle']
);
```
?

Issue #237, opened 2017-06-02
CHANGELOG for components?


manatarms:
Just throwing out an idea since we're all systems go on Fela migrations.
Can we start adding CHANGELOG files to individual packages when they're published. Simple stuff like what was upgraded and why it could break.

I've been trying to migrate a few things in www-next but not knowing what could have broken is a big pain. I'm just guessing and hoping that nothing has broken. For example, `cf-component-heading` says it's a breaking change but nothing seems to be broken. (Which scares me).

What do you'll think?

hturan:
We really, _really_ need this.

koddsson:
>For example, cf-component-heading says it's a breaking change but nothing seems to be broken.

I make breaking changes whenever the API changes to communicate to consumers of this library that their code might not work any more and I think I did it that way this time around I think. Breaking change doesn't mean anything is broken but is more that there is a possibility of something breaking when you update.

In any case I'm down for this. `lerna` has a [conventional commits flag](https://github.com/lerna/lerna#--conventional-commits) that allows us to generate a CHANGELOG automatically that I think would help a lot. I also determines the version bumps from those commits so I think the benefits are great.

manatarms:
I was trying to say that I had no idea if something could break or not. I'm don't deny that Fela stuff should definitely a breaking change but it would just be nice to see what was changed. :)

koddsson:
True, I've been going into cf-ui and running the following to get a diff of the changes.

```
git diff cf-component-button@2.1.2..cf-component-button@4.1.5 -- packages/cf-component-button
```

It's not perfect but it helps. In any case I think the conventional commits would be super helpful. Takes away a lot of pain without much work.

I'll add it to the weekly meetings if someone minds if we enforce this.

manatarms:
I've looked at most button usage. Trying to get LB and RL to work with dev so I check out those as well.

PR #269, opened 2017-06-22
chore(tooling): force conventional commits

koddsson:
Add the packages `husky` and `validate-commit-msg` to create a precommit
git hook that rejects commits that do not adhere to the conventional
commits standard[1].

This enables us to run `lerna publish` with the flag
`--conventional-commits` which will autogenerate a changelog and set the
package versions automatically. This PR enables this flag.

[1]: https://conventionalcommits.org/

Closes: #237

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,17 @@ yarn run bootstrap
 > `yarn run bootstrap` anytime that you have changed the dependencies within a
 > package.
 
+## Commiting changes
+
+`cf-ui` follows the [`conventional-commits`](http://conventionalcommits.org/)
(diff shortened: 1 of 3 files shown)


sejoker:
@koddsson could you update CONTRIBUTING.md with link to https://conventionalcommits.org/ and small explanation why we decided to go with it.

koddsson:
@sejoker Sure! I was thinking about also not merging this until next monday so I can walk through what this change would mean for the team?

sejoker:
@koddsson cool thanks, that would be nice.

koddsson:
I tried this on a different branch and it works like a charm! Didn't prompt me for the change but determined it from the commit.

image

PR #270, opened 2017-06-22
Replace merge-options with seamless-immutable

wyuenho:
merge-options 1.0.0 has a package.json file that just assumes the environment will only be node, this PR swaps it out with seamless-immutable, so the themes merged now are immutable and deeply cloned.

### DIFF
--- a/packages/cf-style-container/package.json
+++ b/packages/cf-style-container/package.json
@@ -15,6 +15,7 @@
     "merge-options": "^1.0.0",
     "prop-types": "^15.5.8",
     "react-fela": "^5.0.0",
+    "seamless-immutable": "^7.1.2",
     "underscore.string": "^3.3.4"
   },
   "peerDependencies": {
--- a/packages/cf-style-container/package.json
+++ b/packages/cf-style-container/package.json
@@ -12,7 +12,6 @@
   },
   "dependencies": {
     "fela": "^5.0.0",
-    "merge-options": "^1.0.0",
     "prop-types": "^15.5.8",
     "react-fela": "^5.0.0",
     "seamless-immutable": "^7.1.2",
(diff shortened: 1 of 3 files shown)


koddsson:
Can we remove `merge-options`?

packages/cf-style-container/test/index.js
```diff
@@ -11,22 +11,33 @@ import { variables } from 'cf-style-const';
...
 
+test('mergeThemes should return an immutable and deeply cloned object', () => {
+  const themeA = () => ({ color: 'yellow' });
+  const themeB = { color: 'blue', breakpoints: { desktop: '1em' } };
+  const props = mergeThemes(variables, themeA, themeB);
+  expect(() => {
+    props.theme.color = 'white';
+  }).toThrow();
+  expect(props.theme.color).toEqual('blue');
+  expect(props.theme.breakpoints).not.toBe(themeB.breakpoints);
```
koddsson: Why is this not the same? Should the `theme` have the breakpoints from theme b?

packages/cf-style-container/package.json
```diff
@@ -15,6 +15,7 @@
     "merge-options": "^1.0.0",
     "prop-types": "^15.5.8",
     "react-fela": "^5.0.0",
+    "seamless-immutable": "^7.1.2",
```
sejoker: it's safe to remove `merge-options`, the package was used in single place.

packages/cf-style-container/test/index.js
wyuenho: `toBe` compares by `===`, I'm deeply cloning the themes, so they should not be identical.

jwineman:
I'm happy we're switching to `seamless-immutable`.  Thanks for debugging @wyuenho.

packages/cf-style-container/test/index.js
koddsson: Ah, reading the test seems to indicate that this shouldn't be the same when in fact is, right? Could we make it more clear that we are testing for "reference equality"? Maybe add a comment or something?

PR #272, opened 2017-06-23
fix(cf-component-form): remove background color

koddsson:
Removes the white form background color.


sejoker:
@koddsson we are talking about Theme value, instead of changing existing Theme (you should have a good reason), just create a new theme and use it. I don't get the idea behind the change.

koddsson:
@sejoker The reason being that setting the background color to white is different to how the component was behaving before the migration to `fela` and breaks things when integrating into cf-www-next.

PR #273, opened 2017-06-23
fix(cf-component-form): remove missing background

koddsson:
We removed the `background` property from the theme 46cbe7b but forgot
to remove it from the component as well.

I realized just as I hit merge on #272  😭


sejoker:
I think we should keep setting background from the theme.

sejoker:
what's wrong with it?

koddsson:
I think you don't want to have a white background all the time but rather a transparent one. Why would forms always be white?

It breaks when implementing the form in load balancing where it expects a transparent background:
image

PR #274, opened 2017-06-23
Removing top padding from page

jculvey:
There's already padding by way of zone-nav, so this adds way too much.
I'm whitelisting it and making it overridable via theme in case it
needs tweaking on some pages.

### DIFF
--- a/packages/cf-component-page/src/Page.js
+++ b/packages/cf-component-page/src/Page.js
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import { createComponent } from 'cf-style-container';
 
 // Doesn't allow overrides atm
-const styles = () => ({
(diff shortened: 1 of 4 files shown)


koddsson:
Could you rebase your commit messages to adhere to the conventional commits standard since I merged #269?

I can even get in here and do it myself if you'd like, just let me know :3

jculvey:
@koddsson done :)

koddsson:
I'm sorry, I probably didn't explain well enough but you need to put the full component name in the scope for reap the automatic versioning benefit. Basically changing the commit message to: `fix(cf-component-page): Removing top padding from page`.

Sorry for not being clear, we might wanna make tooling better on this and I've got some thoughts on how to improve it in the future :)

PR #276, opened 2017-06-25
Format media query snapshots and upgrade polished because styled-comp…

wyuenho:
…onents/polished#140

This PR fixes 2 issues:

1. #271 introduced style snapshotting but Fela renderer's formatting for media query isn't ideal, this PR improves on it by using clean-css to format the styles Fela generates.

2. styled-components/polished#140 is causing Fela to generate `content: ;` for clear fix, clean-css can't parse it and obviously the styles sent down the client is also wrong. Fixed.


sejoker:
make sense to move `felaTestContext` and `felaSnapshot` into dedicated cf-test-something package

wyuenho:
Agree, let's deal with that later

sejoker:
@wyuenho please update `cf-style-provider` README according to the latest changes with direct import and default prefix.

wyuenho:
It's already done here https://github.com/cloudflare/cf-ui/pull/276/files#diff-091a514c8b5c169c735b5f62166f7dae

PR #277, opened 2017-06-27
cf-style-provider update

tajo:
- remove selectorPrefix (again) since it's been fixed in Fela
- replace `clean-css` dep by a regexp
- export felaTestContext and felaSnapshot out of cf-style-provider (again) since it works
- move js-beautify dep to root package.json since we don't need it in cf-style-provider deps

I've tested it in cf-www-next and both dev and prod builds are ok.

### DIFF
--- a/packages/cf-style-provider/package.json
+++ b/packages/cf-style-provider/package.json
@@ -13,11 +13,9 @@
   "dependencies": {
     "cf-style-const": "^1.6.2",
     "cf-style-container": "^1.3.2",
-    "clean-css": "^4.1.4",
-    "cssbeautify": "^0.3.1",
     "fela": "^5.0.0",
     "fela-beautifier": "^5.0.0",
-    "fela-monolithic": "^5.0.0",
+    "fela-monolithic": "^5.0.2",
     "fela-plugin-embedded": "^5.0.0",
     "fela-plugin-fallback-value": "^5.0.0",
     "fela-plugin-lvha": "^5.0.0",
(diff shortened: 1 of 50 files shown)


tajo:
Hopefully this should be the last big update of all snapshots for now.

packages/cf-style-provider/src/felaSnapshot.js
```diff
@@ -1,20 +1,28 @@
...
 
+  const prettifyFelaString = str =>
+    str
+      // adds extra lines before @media
+      .replace(/\@media/g, '\n\n@media')
+      // adds extra lines between keyframes and following selectors
+      .replace(/\}\.([a-zA-Z])/g, '}\n\n.$1')
+      // adds extra indentation and lines to nested selectors
+      .replace(
+        /\)\{(\.[\s\S]+?\})(\})/g,
```
wyuenho: This will only work as long as Fela's formatting remains the same. Fragile as it is, it does generate some pretty good media query formats. Looks good to me.

packages/cf-style-provider/package.json
```diff
@@ -13,11 +13,9 @@
   "dependencies": {
     "cf-style-const": "^1.6.2",
     "cf-style-container": "^1.3.2",
-    "clean-css": "^4.1.4",
-    "cssbeautify": "^0.3.1",
     "fela": "^5.0.0",
     "fela-beautifier": "^5.0.0",
-    "fela-monolithic": "^5.0.0",
+    "fela-monolithic": "^5.0.2",
```
wyuenho: Were there any new fela* packages that we can also do a drive by upgrade?

packages/cf-style-provider/package.json
```diff
@@ -13,11 +13,9 @@
   "dependencies": {
     "cf-style-const": "^1.6.2",
     "cf-style-container": "^1.3.2",
-    "clean-css": "^4.1.4",
-    "cssbeautify": "^0.3.1",
```
wyuenho: Good catch

packages/cf-style-provider/package.json
tajo: Nope and this is a precaution anyway since fela* versions are not really locked.

Issue #234, opened 2017-05-31
`cf-component-label` throws an error


koddsson:
See: https://www.webpackbin.com/bins/-KlSz31gitKKN_SLzWkp for reproduction.

koddsson:
This is actually a integration error that's not super apperent. I think it's just missing the `fela` context.

wyuenho:
I think this is very surprising to me as well. I think we are moving towards providing both unstyled and styled components. cf-component-label, but in reality, unstyled components are still styled, they are just "unthemed", and should be called as such.

A related issue is, I think we should also provide completely unstyled components. I'll file a ticket for that later to lay out the reason.

tajo:
> This is actually a integration error that's not super apperent. I think it's just missing the fela context.

Not a bug. It's missing the base theme. You would have to add `<ThemeProvider>` and pass down `cf-style-const/variables` to make your example working (also you missed required `type` prop for the label). Or just can [just use](https://www.webpackbin.com/bins/-KnGLnwBodTSvrVAJzmo) `cf-style-provider/<StyleProvider>` that does all that.

PR #278, opened 2017-06-27
Stop passing rule functions downstream

tajo:
This prevents to append custom classNames to components styled by Fela.

It gives a warning: 

```
You cannot restyle Fela component ${componentName}
```

### DIFF
--- a/packages/cf-style-container/package.json
+++ b/packages/cf-style-container/package.json
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "fela": "^5.0.0",
+    "fela-utils": "^5.0.2",
     "prop-types": "^15.5.8",
     "react-fela": "^5.0.0",
     "seamless-immutable": "^7.1.2",
--- a/packages/cf-style-container/src/createComponent.js
+++ b/packages/cf-style-container/src/createComponent.js
@@ -1,0 +1,72 @@
+import { extractPassThroughProps, resolvePassThrough } from 'fela-utils';
+import { createElement } from 'react';
+import PropTypes from 'prop-types';
+
+function createComponentFactory(createElement, contextTypes) {
+  return function createComponent(rule, type, passThroughProps) {
+    const displayName = rule.name ? rule.name : 'FelaComponent';
+
+    const FelaComponent = ({ children, ...ruleProps }, { renderer, theme }) => {
+      if (!renderer) {
+        throw new Error(
+          "createComponent() can't render styles without the renderer in the context. Missing react-fela's <Provider /> at the app root?"
+        );
+      }
+
+      const componentName = typeof type === 'string'
+        ? type
+        : type.displayName || type.name || '';
+
+      // compose passThrough props from arrays or functions
+      const resolvedPassThrough = [
+        ...resolvePassThrough(passThroughProps, ruleProps)
+      ];
+
+      if (ruleProps.className) {
+        console.warn(`You cannot restyle Fela component ${componentName}`);
+      }
+
+      const componentProps = extractPassThroughProps(
+        resolvedPassThrough,
+        ruleProps
+      );
+
+      ruleProps.theme = theme || {};
+
+      if (ruleProps.style) {
+        componentProps.style = ruleProps.style;
+      }
+
+      if (ruleProps.className) {
+      }
+
+      componentProps.className = renderer.renderRule(rule, ruleProps);
+
+      if (ruleProps.id) {
+        componentProps.id = ruleProps.id;
+      }
+
+      if (ruleProps.innerRef) {
+        componentProps.ref = ruleProps.innerRef;
+      }
+
+      const customType = ruleProps.is || type;
+      return createElement(customType, componentProps, children);
+    };
+
+    if (contextTypes) {
+      FelaComponent.contextTypes = contextTypes;
+    }
+
+    // use the rule name as display name to better debug with react inspector
+    FelaComponent.displayName = displayName;
+    FelaComponent._isFelaComponent = true;
+
+    return FelaComponent;
+  };
+}
+
+export default createComponentFactory(createElement, {
+  renderer: PropTypes.object,
+  theme: PropTypes.object
+});
--- a/packages/cf-style-container/src/createComponent.js
+++ b/packages/cf-style-container/src/createComponent.js
@@ -4,7 +4,9 @@ import PropTypes from 'prop-types';
 
 function createComponentFactory(createElement, contextTypes) {
   return function createComponent(rule, type, passThroughProps) {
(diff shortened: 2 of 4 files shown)


packages/cf-style-container/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "fela": "^5.0.0",
+    "fela-utils": "^5.0.2",
```
marksteyn: The readme for fela-utils says "fela-utils is a private utility package for Fela and should only be used interally."

packages/cf-style-container/src/createComponent.js
```diff
@@ -0,0 +1,72 @@
...
+import PropTypes from 'prop-types';
+
+function createComponentFactory(createElement, contextTypes) {
+  return function createComponent(rule, type, passThroughProps) {
+    const displayName = rule.name ? rule.name : 'FelaComponent';
+
+    const FelaComponent = ({ children, ...ruleProps }, { renderer, theme }) => {
+      if (!renderer) {
+        throw new Error(
+          "createComponent() can't render styles without the renderer in the context. Missing react-fela's <Provider /> at the app root?"
```
koddsson: Should this be [`ThemeProvider`](https://github.com/cloudflare/cf-ui/issues/234#issuecomment-310485817)?

packages/cf-style-container/src/createComponent.js
```diff
@@ -0,0 +1,72 @@
+import { extractPassThroughProps, resolvePassThrough } from 'fela-utils';
+import { createElement } from 'react';
+import PropTypes from 'prop-types';
+
+function createComponentFactory(createElement, contextTypes) {
+  return function createComponent(rule, type, passThroughProps) {
+    const displayName = rule.name ? rule.name : 'FelaComponent';
```
koddsson: Would this be simpler as `const displayName = rule.name || 'FelaComponent'`?

Could we also add a comment about when `rule.name` would be falsy? I can imagine someone seeing a lot of `FelaComponent`s in their code and wanting to know why, digging through the code to find this line so it would be good to have a comment explaining why it's happening?

packages/cf-style-container/src/createComponent.js
```diff
@@ -0,0 +1,72 @@
...
+        ruleProps
+      );
+
+      ruleProps.theme = theme || {};
+
+      if (ruleProps.style) {
+        componentProps.style = ruleProps.style;
+      }
+
+      if (ruleProps.className) {
```
koddsson: Remove this?

koddsson:
Nice conventional commit 👌 🎉

wyuenho:
So I see you've basically just inlined `createComponentFactory` from [here](https://github.com/rofrischmann/fela/blob/master/packages/fela/src/bindings/createComponentFactory.js), with a few minor changes.

I admire this attempt to solve this [problem](https://github.com/cloudflare/cf-ui/pull/256#discussion_r124151842), but this approach falls far shorter than what's necessary to do what you want to achieve.

I can still do this:

```javascript
import { createComponent } from 'react-fela';
import { Button } from 'cf-component-button';
const MyButton = createComponent(() => ({background: 'black'}), Button, Object.keys(Button.propTypes));
```
This is still not hard at all to do at all, and easy to miss when reviewing PRs.

An additional problem with this approach is, fela doesn't have to tell us when they violate semver and break the internal machinery of fela-utils, I'm very hesitant to use something fela explicitly said you should only use it for fela internals.

What you can possibly do is to monkey-patch fela in our application, but I'm not sure that will pass code review.

packages/cf-style-container/src/createComponent.js
tajo: No, it's correct.

packages/cf-style-container/src/createComponent.js
tajo: Good catch. Fixed.

packages/cf-style-container/src/createComponent.js
tajo: Fixed. It's falsy if you use anonymous function as

```js
const Foo = createComponent(() => { my_rule}, 'h1')
```

vs 

```js
const FooStyle = () => { my_rule};
const Foo = createComponent(FooStyle, 'h1')
```

It's like other javascripts. So far we are mostly using that first shorter syntax and doesn't really cause any troubles with debugging I think.

packages/cf-style-container/package.json
tajo: It's fine. I helped creating that package. It's using semver.

packages/cf-style-container/src/createComponent.js
koddsson: But won't that then crash like in [this bug report](https://github.com/cloudflare/cf-ui/issues/234#issuecomment-310485817) since it won't have a base theme?

tajo:
@wyuenho 

> So I see you've basically just inlined createComponentFactory from here, with a few minor changes.

Yep. I made it simpler.

I don't think that your example really works and even if it did, `react-fela` will soon not be a dependency of any of our projects. Also, I am more than happy to enforce that through eslint rule (together with not importing cf-style-const and other fela packages)  if people try to add that.

> fela doesn't have to tell us when they violate semver

It will.

tajo:
@koddsson It can't be pushed upstream because Fela's implementation is more permissive and cover more cases which make sense (for example it needs to play nicely with legacy CSS and classNames or support for React Native).

wyuenho:
Okay, banning importing from fela via an eslint rule will make it sufficiently harder for me to override styles.

wyuenho:
> I don't think that your example really works and even if it did, react-fela will soon not be a dependency of any of our projects. 

Are we just going to wrap everything react-fela exports? That sounds good to me too.

packages/cf-style-container/src/createComponent.js
tajo: Those are two different things. `createComponent` should crash if there's missing `renderer` in the context.

The other bug is because our components have hard assumption that our base theme (cf-style-const) is in the context (ThemeProvider puts it into that context or in our case StyleProvider). I used to have a warning for that but it had to be in every component, which was annoying. **Now, I just put it here as an another error message.**

tajo:
> Are we just going to wrap everything react-fela exports? That sounds good to me too.

Yes. We already do it. I just needed to update `<StyleProvider>` (merged today) so it accepts outside `renderer` and works in next.

tajo:
Added another commit that fixes propType warnings. It passes propTypes to the outer component, so it works and uses component name for displayName instead of style function name (which is often anonymous or not useful anyway).

```js
<Button />
```

gives

screen shot 2017-06-27 at 10 32 35 am

Issue #228, opened 2017-05-26
Migrate `cf-component-toggle` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

Issue #239, opened 2017-06-05
Fixed HEX codes in `cf-style-const`


koddsson:
These dynamic hex values in `cf-style-const` make it extremely hard to find the color you want, we should just use the hex values.

wyuenho:
That's what a style guide is for.

koddsson:
What styleguide? https://cloudflare.github.io/cf-ui/ ?

wyuenho:
No, that's an API doc inappropriately named to styleguide, a styleguide should show case the color palette, design details such as the grid system used, fonts, and sizes, illustrated by common examples of interactions among components when composed. Basically, whatever @hturan was working on.

koddsson:
I don't think @hturan is working on anything in any official capacity (he can correct me if I'm wrong).

I think I'd much more prefer just not using the `darken(...)/lighten(...)` functions like we do now so that it's more clear what colors are declared.

wyuenho:
Just curious, under what circumstances will you want to find a color?

koddsson:
Sampling from other parts of the system or design and trying to find that color in `cf-style-const`.

hturan:
> I don't think @hturan is working on anything in any official capacity (he can correct me if I'm wrong).

That's correct (although I wish it wasn't)!

---

I think this is more of a documentation issue than an implementation one.

If there was a page you could visit that visually displayed our colours, that would solve this (I think). It doesn't really matter if we implement constants as a hex-code string, an`rgba` value, or a result of a function call, since it's always going to be a valid CSS colour.

Let's add a `cf-style-const` (or something) section to our cf-ui docs that documents the package contents in a visual manner.

wyuenho:
> If there was a page you could visit that visually displayed our colours, that would solve this (I think). 

Yes. I think I was referring to the styleguide-audit project you were working on in unofficial capacity

koddsson:
So move this to the `cf-styleguide` project?

koddsson:
What's going to happen here? Is someone going to add documentation to `cf-styleguide`?

manatarms:
I'm working on adding documentation to cf-styleguide for colors and icons.

wyuenho:
@mrmrs You may want to coordinate with the people here.

mrmrs:
@manatarms Where is the info coming from that you are adding to the documentation re: colors?

manatarms:
@mrmrs PMed you for extended convo.

koddsson:
[...]

PR #281, opened 2017-06-28
feat(cf-component-toggle): rewrite in `fela`

koddsson:
Now `cf-component-toggle` uses `fela` to apply styles defined in JS to
the components rather than relying on outside CSS to provide styles.

BREAKING CHANGE: `className` prop is no longer applied to the component.

Closes: #228

### DIFF
--- a/packages/cf-component-toggle/src/Toggle.js
+++ b/packages/cf-component-toggle/src/Toggle.js
@@ -1,6 +1,115 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme, disabled }) => {
+  return {
+    display: 'block',
+    position: 'relative',
+    margin: 0,
+    height: theme.height,
+    width: theme.width,
+    borderRadius: theme.borderRadius,
+    borderWidth: theme.borderWidth,
+    borderStyle: theme.borderStyle,
+    borderColor: disabled ? theme.colorGray : theme.borderColor,
+    cursor: disabled ? 'not-allowed' : 'pointer',
+    fontSize: theme.fontSize,
+    marginBottom: theme.marginBottom,
+    minHeight: theme.minHeight,
+
+    '&::before': {
+      display: 'block',
+      position: 'absolute',
+      top: 0,
+      bottom: 0,
+      width: '50%',
+      lineHeight: theme['&::before'].lineHeight,
+      textAlign: 'center',
+      content: "'On'",
+      left: 0,
+      background: disabled ? theme.colorGray : theme['&::before'].background,
+      color: disabled ? theme.colorGrayLight : theme['&::before'].color
+    },
+    '&::after': {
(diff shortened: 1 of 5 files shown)


koddsson:
Forgot the tests 😓

packages/cf-component-toggle/src/Toggle.js
```diff
@@ -58,4 +156,4 @@ Toggle.propTypes = {
   disabled: PropTypes.bool
```
sejoker: please add className into propTypes

packages/cf-component-toggle/src/ToggleTheme.js
```diff
@@ -0,0 +1,46 @@
+export default baseTheme => ({
+  display: 'block',
+  position: 'relative',
+  margin: 0,
+  height: '2.26667rem',
+  width: '5.334rem',
+  borderRadius: '2px',
+  borderWidth: '1px',
+  borderStyle: 'solid',
+  borderColor: '#aaa',
```
sejoker: one-off color?

koddsson: I dunno, I took it directly from the CSS and [am unsure on how to find a equivilent one](https://github.com/cloudflare/cf-ui/issues/239).

manatarms: I think @Neynah can  help you with the colors

koddsson:
>all values hardcoded in the component `styles`

Didn't do this because I was 100% sure someone would point to that in my PR and reject it on the basis that it would have hardcoded values. I might just refer back to this comment if I end up doing it that way in the future.

packages/cf-component-toggle/src/ToggleTheme.js
koddsson: I don't wanna wait forever for something to happen here so I'm gonna merge this with my two approvals.

If someone wants to make a issue and assign it to me with what they would like to happen here I will try to schedule it to be worked on as soon as possible.

hturan:
Did we miss something here? Internally, `toggle.css` has:

```sass
input[type="checkbox"].cf-toggle__checkbox,
.cf-toggle__checkbox,
.cf-toggle__label {
  @extend %cf-a11y-sr_only;
}
```

Yet in this PR we're only applying `cf-a11y-sr_only` rules to `A11yLabel` (if I'm reading things right).

koddsson:
@hturan, yep. Check out #284.

koddsson:
Here's a even better link: https://github.com/cloudflare/cf-ui/pull/284/files#diff-62da98b021f4974133c4763ef647d637

hturan:
@koddsson Ah cool, sorry! That clarifies things.

koddsson:
No worries :) I often miss things in PRs in cf-ui not realizing it until I implement it in the dashboard. The integration we currently have isn't the best at surfacing these issues before integrating :(

PR #282, opened 2017-06-28
Remove --independent flag

koddsson:
We are already in independant mode, see `lerna.json` in the project root.


koddsson:
Aaah, it's not a conventional commit! But I guess it doesn't matter since it's not in one of the `packages/*` folders.

PR #283, opened 2017-06-28
Publish design docs

manatarms:
Creating new design docs.


tajo:
Meh, just push it directly to master the next time.

manatarms:
Okie :D

koddsson:
>Meh, just push it directly to master the next time.

Is this a road we want to go down? [I've thought about it previously](https://github.com/cloudflare/cf-ui/pull/282) but I felt like it would open us up to disagreements and unpleasantries as well as putting the decision on what needs to go to a PR on individuals where their opinion on what can be directly pushed varies.

tajo:
> Is this a road we want to go down?

For sure. This is same as publishing cf-ui. It's also pushed directly to the master. There's no value in reviewing unreadable output from webpack. Ideally this should be more automatic (auto publish after merge into cf-ui, cf-ux or cf-styleguide).

PR #284, opened 2017-06-29
Hide checkbox in toggle

koddsson:
When applying the new version of `cf-component-toggle` in the dashboard it has some extra styles from the CSS that is being applied:

image

This PR adds the same a11y "hide" function to the checkbox (like how it was done previously in the CSS) as well as introducing the !important plugin so that the CSS doesn't overwrite those styles.


tajo:
Appending `!important` to every single rule generated by Fela in all our projects can be a risky change. However, I think it's the right thing to do (I even proposed to create that plugin). Some [other](https://github.com/Khan/aphrodite#disabling-important) CSS in JS libs do it by default.

Any thoughts from others?

wyuenho:
This is I think the 4th time I see the checkbox showing up in a PR. In terms of hygiene, I'd much rather see where the bleed was coming from instead of using the nuclear option, however, if it's too hard to fix the bleed, using `!important` everywhere *just on this component* is fine. I'm still not particularly keen on suffixing `!important` everywhere. One you go down that path, your CSS size effectively doubles (or more), and you give up on an incentive for you to look for style bleeds, and fix it locally with more efficient means, such as just overriding a few properties.

koddsson:
I don't think [`fela-plugin-important`](https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-important/README.md) offers that we only put it on one component.

I'm open to suggestions but this seems like the canonical way to do this in `fela` right now so I went that route.

mrmrs:
Take all of this with a large grain of salt as I am not managing this code base at all. Based on my experience setting important on everything is a significant code smell.  I've never had to do this in React while managing styles with javascript. Generally have handled it by having an HOC that prevents style bleed by setting the desired default values. !important will not solve this across the board here as it can still be overridden by a variety of selectors that also use !important unless I'm missing something from the implementation details.

Alternative solutions that in my past experience are a bit more bullet proof for this type of thing:

- Dynamically set the specificity of the outputted css selectors from fela based on heaviest weighted selector in the codebase 
- Implement a scoping component that can add a parent classname with a desired specificity level
- Use inline styles (these trump anything in css but are also slower to render)

hturan:
> !important will not solve this across the board here as it can still be overridden by a variety of selectors that also use !important unless I'm missing something from the implementation details.

You're not, that is correct.

For context:

The situation here is that `fela` gives us a class name (e.g. `.cf-abcdef`) that is out-specified by something else in our codebase (e.g. `input[type=checkbox]`). `.cf-abcdef` has a specificity of `0,0,1,0` and `input[type=checkbox]` has a specificity of `0,0,1,1`, so it wins out.

The issue here is that if the rules defined in `input[type=checkbox]` have `!important`, they will still override what is specified within `fela`.

koddsson:
All good comments; I'm really not trying to solve the issue across the board here and would expect developers that are migrating components to test their components to make sure they don't behave differently after migrating to `fela`.

If we want to try to solve this across the board that's a bigger issue that we can tackle as part of this PR and if the general consensus is to do that I think we should try to list up the different ways styles can bleed into a component and work our way to a solution. I suspect if we want to solve this across the board we are going to need to make some significant changes to `fela` or write a different plugin to set a more specific CSS selector.

I'm personally leaning towards keeping this as-is until we migrate all the components and remove the global stylesheets. There are not a ton of `!important` keywords in the dashboard so I feel like the impact is minimal (but I could be wrong).

wyuenho:
I think the point is, override is a fact of life, once there's a `!important` introduced, whenever we need to override a few of these styles will become a war of `!important`. I'd much rather see at least we've ID'ed the originating offending styles, perhaps even fix those in our application code base or cf-ui. It's not that hard, its keeps the style size more or less stable, and 1 less plugin to manage.

koddsson:
@wyuenho Sure, I can do that.

koddsson:
The offending CSS looks something like this: 

```css
input[type="radio"],input[type="checkbox"] {
    height: 15px;
    line-height: normal;
    margin: 1px 0 0;
    padding: 0;
    position: relative;
    width: 15px;
    -moz-appearance: none;
    -webkit-appearance: none;
}
```

Resulting in something like this:

image

Maybe we could add a plugin or make a PR to `fela` to add the tag name before the classname to get a more specific selector, but I feel like we're always going to have this problem of which selector has precedence.

It would be interesting to try to leverage the shadow dom if it [was better supported](https://caniuse.com/#feat=shadowdom).

One thing that comes to mind is to change this `input[type="checkbox"]` rule to a class name and update the dashboard code to add this class name to all the inputs. I feel like that may solve this issue in a very clean way. What do you think?

wyuenho:
Can we use `createComponent` to put a `display: none` or something to the checkbox?

https://github.com/cloudflare/cf-ui/blob/master/packages/cf-component-toggle/src/Toggle.js#L131

koddsson:
Yup, I considered that but I figured the reason why we weren't doing it was because of accessibility.

wyuenho:
This is the effective SASS style now:

screen shot 2017-07-04 at 18 09 42

Just put a negative margin on it.

koddsson:
Sure. If that's cool with everyone. I'm not super sure how accessibility should work so I wasn't sure what to do here.

koddsson:
Looking at this better I don't think that margin would help? If you're thinking about moving the out of the possible scope of the page I think the `left` property would work better.

koddsson:
[This guide](http://www.456bereastreet.com/archive/201211/accessible_custom_checkboxes_and_radio_buttons/) seems to indicate that we can hide the input with `opacity: 0` so I'm gonna do that.

koddsson:
How's this @mrmrs @wyuenho @tajo @hturan ?

sejoker:
much simpler solution with opacity, go for it.

packages/cf-style-provider/package.json
```diff
@@ -31,4 +31,4 @@
   "peerDependencies": {
     "react": "^0.14.9 || ^15.0.0-0 || ^16.0.0-0"
   }
-}
+}
```
tajo: You are going to hate me but new lines at the EOF are a good [thing](https://evanhahn.com/newline-necessary-at-the-end-of-javascript-files/). 😆

koddsson: My editor normally does that, not sure what happened here.

koddsson: Really feels like it should happen automatically with some thing like [`eol-last`](http://eslint.org/docs/rules/eol-last). No need for a human to add them in manually.

koddsson: Oh, this is a JSON file and not a javascript file. Having a newline at EOF doesn't apply here since we'll never concat JSON files.

I knew my editor wouldn't betray me!

manatarms: Which editor may I ask?

koddsson: Visual Studio Code of course. Highly recommend.

manatarms: Hmmm. I'll give it a try. But I'm really attached to my sublime config

koddsson: No worries, I really like VSCODE but if sublime is your jam you should just stick with what you like :)

Issue #230, opened 2017-05-26
Migrate `cf-component-progress` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

PR #285, opened 2017-06-29
feat(cf-component-progress): implement in fela

koddsson:
BREAKING CHANGE: Underlying styling have been completely re-written in `fela` warranting extra caution when updating as well as a major version bump.

Closes: #230

### DIFF
--- a/packages/cf-component-progress/src/Progress.js
+++ b/packages/cf-component-progress/src/Progress.js
@@ -1,6 +1,92 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Link from 'cf-component-link';
+import { createComponent } from 'cf-style-container';
+import { clearFix } from 'polished';
+
+const styles = () => ({
+  position: 'relative'
+});
+
+const Bar = createComponent(
+  ({ theme }) => ({
+    display: 'block',
+    width: '100%',
+    appearance: 'none',
+
+    '&::-webkit-progress-bar': {
+      backgroundColor: theme.bodyBackground,
+      color: theme.bodyBackground
+    },
+
+    '&::-webkit-progress-value': {
+      backgroundColor: theme.color.marine,
+      color: theme.color.marine,
+      transition: 'width 500ms ease'
+    },
+
+    /* Mozilla uses progress-bar as the value */
+    '&::-moz-progress-bar': {
+      backgroundColor: theme.color.marine,
+      color: theme.color.marine
+    }
+  }),
+  'progress',
+  ['max', 'value']
+);
+
+const Items = createComponent(
+  () => ({
+    listStyle: 'none',
+    ...clearFix(),
+    margin: 0,
+    padding: 0
+  }),
+  'ol'
+);
+
+const Item = createComponent(
+  ({ theme, width, disabled, active }) => {
+    let color = theme.colorGray;
+    if (disabled) {
+      color = theme.colorGrayLight;
+    } else if (active) {
+      color = theme.color.marine;
+    }
+
+    return {
+      width,
+      float: 'left',
+      height: 'auto',
+      padding: 0,
+      paddingTop: '7.5px',
+
+      color,
+      fontSize: '0.86667em',
+
+      textAlign: 'center',
+      cursor: disabled ? 'default' : 'pointer',
+
+      '&::before': {
+        content: 'none'
+      },
+
+      tablet: {
+        display: active ? 'block' : 'none',
+        width: 'auto !important',
+        float: 'none'
+      },
+
+      '& > .cf-link': {
+        display: 'block',
+        color: 'inherit',
+        cursor: 'pointer'
+      }
+    };
+  },
+  'li',
+  ['key', 'width', 'disabled', 'active']
+);
 
 class Progress extends React.Component {
   constructor(props) {
@@ -15,41 +101,42 @@ class Progress extends React.Component {
   }
 
   render() {
-    const max = this.props.steps.length;
+    const { className, steps, active } = this.props;
+    const max = steps.length;
     const itemWidth = (1 / max * 100).toFixed(4) + '%';
 
     let value;
 
-    const items = this.props.steps.map((step, index) => {
-      let className = 'cf-progress__item';
+    const items = steps.map((step, index) => {
+      const isActive = step.id === active;
 
-      if (step.id === this.props.active) {
+      if (isActive) {
         value = index + 1;
-        className += ' cf-progress__item--active';
-      }
-
-      if (step.disabled) {
-        className += ' cf-progress__item--disabled';
       }
 
       return (
-        <li key={step.id} className={className} style={{ width: itemWidth }}>
+        <Item
+          key={step.id}
+          width={itemWidth}
+          disabled={step.disabled}
+          active={isActive}
+        >
           <Link
             onClick={this.handleClick.bind(null, step)}
             disabled={step.disabled}
           >
             {step.label}
           </Link>
-        </li>
+        </Item>
       );
     });
 
     return (
-      <div className="cf-progress">
-        <progress className="cf-progress__bar" max={max} value={value} />
-        <ol className="cf-progress__items">
+      <div className={className}>
+        <Bar max={max} value={value} />
+        <Items>
           {items}
-        </ol>
+        </Items>
       </div>
     );
   }
@@ -64,11 +151,12 @@ Progress.propTypes = {
       label: PropTypes.string.isRequired,
       disabled: PropTypes.bool.isRequired
     })
-  ).isRequired
+  ).isRequired,
+  className: PropTypes.string
 };
 
 Progress.defaultProps = {
   steps: []
 };
 
-export default Progress;
+export default createComponent(styles, Progress);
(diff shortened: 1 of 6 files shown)


packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+      fontSize: '0.86667em',
+
+      textAlign: 'center',
+      cursor: disabled ? 'default' : 'pointer',
+
+      '&::before': {
+        content: 'none'
+      },
+
+      [`@media (max-width: ${theme.breakpoints.tablet})`]: {
```
tajo: you can replace this just with

```
tablet: {

}
```

packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+        display: active ? 'block' : 'none',
+        width: 'auto !important',
+        float: 'none'
+      },
+
+      '& > .cf-link': {
```
tajo: This will not work when cf-component-link is migrated. The best solution here is to avoid the child selector, add some extra prop to Link as `isInProgressBar` and put these styles there.

packages/cf-component-progress/src/index.js
```diff
@@ -1,3 +1,6 @@
-import Progress from './Progress';
+import ProgressUnstyled from './Progress';
+import ProgressTheme from './ProgressTheme';
 
-export default Progress;
+const Progress = applyTheme(ProgressUnstyled, ProgressTheme);
+
+export { Progress as default, ProgressUnstyled, ProgressTheme };
```
tajo: just

```
export { Progress, ProgressUnstyled, ProgressTheme };
```

We never export component as a default.

packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+});
+
+const Bar = createComponent(
+  () => ({
+    display: 'block',
+    width: '100%',
+    appearance: 'none',
+
+    '&::-webkit-progress-bar': {
+      backgroundColor: '#ebebeb',
```
tajo: This color exists in the base theme.

packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+
+    '&::-webkit-progress-value': {
+      backgroundColor: '#2f7bbf',
+      color: '#2f7bbf',
+      transition: 'width 500ms ease'
+    },
+
+    /* Mozilla uses progress-bar as the value */
+    '&::-moz-progress-bar': {
+      backgroundColor: '#2f7bbf',
```
tajo: I would ask designers if we should replace this color by one from the consts. If not, it should be a new color since it's used in  multiple places.

packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+  }),
+  'progress',
+  ['max', 'value']
+);
+
+const Items = createComponent(
+  () => ({
+    listStyle: 'none',
+    '&::after': {
+      content: '',
```
tajo: `content:'""'` [why](http://fela.js.org/docs/introduction/Drawbacks.html)

koddsson: This is super easy to miss. I've done this very recently where I caught this before shipping but I wonder if there's benefit in doing something like this automatically.

tajo: Eventually there will be an eslint plugin for stuff like this.

packages/cf-component-progress/src/index.js
koddsson: Well we do, both in the non-fela packages as well as in other ones such as [Toggle](https://github.com/cloudflare/cf-ui/blob/a49ebb5f87b31209b0beadada966fa8e8d3ed2bf/packages/cf-component-toggle/src/index.js).

Doing it this way means it doesn't break the "import API" so you don't need to update all the places where it's imported in the dashboard.

I'm fine if we want to not use the default keyword but there's very much a precedent to do it like this.

packages/cf-component-progress/src/Progress.js
koddsson: I posed in the design chat.

packages/cf-component-progress/src/index.js
tajo: Oh, the toggle slipped through. All other migrated components don't export default afaik. I think it's better to grep&replace it in our codebase and keep it unified across the library so we don't have to go to documentation every single time to see how to import the component. It's going to be a major release anyway.

Sooner or later you might add another component into this package and then you have a problem because it's not obvious what the default is (think about having Button and a few months later adding ButtonGroup).

koddsson: Yeah I agree, I'm not a fan of the default keyword at all but I figured it would make the migration easier.

packages/cf-component-progress/src/Progress.js
neynah: This color is already living in our color constants- it's marine: https://cloudflare.github.io/cf-ui/#cf-design-colors. If any other colors come up that we should name let us know. :)

koddsson: @neynah That section seems to be missing quite a few colors. Are all those added manually?

packages/cf-component-progress/src/Progress.js
koddsson: I think that's a weird pattern to go down. I think I'll preferably just migrate the link component first.

koddsson: I've opened up #288 to migrate the link component. Will update this PR once that lands.

packages/cf-component-progress/src/Progress.js
tajo: yea, those are generated out of theme.color... so some old colors are missing

koddsson: So all new colors should go to the `theme.color` object? Because I know for a fact that we've been adding colors as `theme.color${colorName}`.

packages/cf-component-progress/src/Progress.js
koddsson: Actually I need to merge this so I can merge `Link` and then I can change this by applying a theme.

packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+  }),
+  'progress',
+  ['max', 'value']
+);
+
+const Items = createComponent(
+  () => ({
+    listStyle: 'none',
+    '&::after': {
+      content: "''",
```
sejoker: better to use instead: https://polished.js.org/docs/#clearfix

packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+  'ol'
+);
+
+const Item = createComponent(
+  ({ theme, width, disabled, active }) => {
+    let color = theme.colorGray;
+    if (disabled) {
+      color = theme.colorGrayLight;
+    } else if (active) {
+      color = '#2f7bbf';
```
sejoker: read from theme.color.marine

packages/cf-component-progress/src/Progress.js
```diff
@@ -1,6 +1,95 @@
...
+      fontSize: '0.86667em',
+
+      textAlign: 'center',
+      cursor: disabled ? 'default' : 'pointer',
+
+      '&::before': {
+        content: 'none'
+      },
+
+      [`@media (max-width: ${theme.breakpoints.tablet})`]: {
```
sejoker: tablet: {
...
}

should work

packages/cf-component-progress/src/ProgressTheme.js
```diff
@@ -0,0 +1 @@
+export default baseTheme => ({});
```
sejoker: I think it's useless unless we have some values

packages/cf-component-progress/src/Progress.js
koddsson: It didn't work previously. Maybe now since #290 has been merged I can rebase and change this.

packages/cf-component-progress/src/ProgressTheme.js
koddsson: https://github.com/cloudflare/cf-ui/pull/281#pullrequestreview-46909882

manatarms: @koddsson Woah! This meta precog stuff just blew my mind.

packages/cf-component-progress/src/index.js
```diff
@@ -1,3 +1,8 @@
-import Progress from './Progress';
+import ProgressUnstyled from './Progress';
+import ProgressTheme from './ProgressTheme';
 
-export default Progress;
+import { applyTheme } from 'cf-style-container';
+
+const Progress = applyTheme(ProgressUnstyled, ProgressTheme);
+
+export { Progress as default, ProgressUnstyled, ProgressTheme };
```
tajo: So can we remove `as default` ?

koddsson: Yup, will do.

Issue #287, opened 2017-06-30
Migrate `cf-component-link` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

See #99 for deprecation talks.

toekneestuck:
Moved internal.

Issue #99, opened 2017-03-16
Deciding the fate of Link component


manatarms:
**Current Situation**
I was migrating ```cf-component-link``` to Fela and noticed that we have a dual implementation in the component. So I have some questions about how we want this component to behave.

Currently, the component accepts a ```tagName``` that is rendered instead of `<a></a>`. The component also accepts a ```to``` prop that is used to route to a location. 

I've noticed use cases of  ```cf-component-link``` where _"link should look like a link but act like a button"_. 

The ```tagName``` and ```to``` props introduce quite a few ```if``` statements into the component. These statements check for ```href```, ```onClick```, ```to``` and call different functions accordingly. This also diverges from the idea of having a standardized component library by allowing a user to render anything as a link using the ```tagName``` prop. 

**Questions**

1. Should we remove the ability to make anything a link? i.e. Always render an ```<a>{children}</a>```
2. Should we remove the ```{to}``` entirely and have just an onClick handler. This will satisfy the button use case. The routing can also then be handled using the onClick handler. 
OR
3. Should we keep the ```{to}``` and remove the ability to call a custom onClick handler. This will force users to restyle the cf-component-button to look like a link (which would now be super easy using Fela). Routing can then be handled using the ```to``` prop.
4. 
![marilee-ullmann-quote-we-love-it-the-way-it-is-its-not-broken-dont-fix](https://cloud.githubusercontent.com/assets/4206893/24019708/5834382e-0a56-11e7-9416-17a7a92193d0.jpeg)

jwineman:
I'm pretty sure we stole the `<Link to={PAGE}/>` interface from React Router.  I'd prefer `<Link onClick={}/>` and have us pass in a bound action creator from react router with the link OR any random function if we want it to behave like a button or whatever.

sejoker:
I think it was designed with this flexibility in mind. What are pain points we have with current implementation?

jwineman:
Implementing `to` couples it to whatever router we're using.  Since `to` and `onChange` both usually end up calling redux actions we can probably consolidate.

manatarms:
With current implementation in order for Fela to accept our onClick handler we need to change the handler before passing it to Fela. Which goes against React philosophy of never changing props.

wyuenho:
[...]

PR #288, opened 2017-06-30
feat(cf-component-link): rewrite styling in fela

koddsson:
BREAKING CHANGE: The `className` is now the one that is passed down from
the `fela` higher order component and any `className` prop that is
passed to `Link` will be ignored.

Closes: #287

### DIFF
--- a/packages/cf-component-link/src/Link.js
+++ b/packages/cf-component-link/src/Link.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { routeTo } from 'cf-util-route-handler';
+import { createComponent } from 'cf-style-container';
 
 class Link extends React.Component {
   constructor(props, context) {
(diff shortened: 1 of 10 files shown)


koddsson:
Aaaaaahhh, there are so many components that depend on `Link`!

packages/cf-component-link/src/Link.js
```diff
@@ -47,22 +48,43 @@ class Link extends React.Component {
...
+          outline: theme['&:focus'].outline,
+          outlineOffset: theme['&:focus'].outlineOffset
+        },
+
+        '&:active': {
+          color: theme['&:active'].color,
+          outline: theme['&:active'].outline
+        }
+      }),
+      tagName,
```
manatarms: Should we be locking this down to a set of acceptable links? Not sure what we decided in #99

koddsson: The component didn't do any kind of validation of the `tagName` before so I'm unsure what these would be. I'd prefer just to have the `a` tag here and not allow it to be customized since I can't see the use case for it.

koddsson:
I'm blocking this until all the component that rely on `Link` have been migrated.

koddsson:
To unblock this the following components need to be migrated to `fela`.

- card
- progress
- dropdown
- link

packages/cf-component-link/src/Link.js
manatarms: I think there is a use case where the link needs to behave like a button or something. I think @jculvey can describe it better than me. But I agree that it should be just an `<a>` tag

packages/cf-component-link/src/Link.js
```diff
@@ -47,22 +48,43 @@ class Link extends React.Component {
...
+      ['disabled', 'href', 'role']
```
tajo: `href` and `role` are missing in proptypes

packages/cf-component-link/src/index.js
```diff
@@ -1,3 +1,11 @@
-import Link from './Link';
+import LinkUnstyled from './Link';
+import LinkTheme from './LinkTheme';
+import { applyTheme } from 'cf-style-container';
 
-export default Link;
+const Link = applyTheme(LinkUnstyled, LinkTheme);
+
+export default {
```
tajo: this should be just `export { Link, LinkUnstyled, LinkTheme }`

koddsson: Yeah I think that's why CI is failing. Once I've merged the other PRs I will come back to this.

Issue #226, opened 2017-05-26
Migrate `cf-component-input` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

PR #289, opened 2017-06-30
feat(cf-component-input): rewrite styling in fela

koddsson:
BREAKING CHANGE: There is no default export in `cf-component-input`
anymore, the previously exported default is now exported as `Input`.

Additionally the `className` prop is now ignored.

Closes: #226

### DIFF
--- a/packages/cf-component-input/example/basic/component.js
+++ b/packages/cf-component-input/example/basic/component.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import Input from 'cf-component-input';
+import { Input } from 'cf-component-input';
 
 class InputComponent extends React.Component {
   constructor(props) {
--- a/packages/cf-component-input/src/Input.js
+++ b/packages/cf-component-input/src/Input.js
@@ -1,22 +1,59 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  zIndex: theme.zIndex,
+  width: theme.width,
+  margin: theme.margin,
+  padding: theme.padding,
+  border: theme.border,
+  borderRadius: theme.borderRadius,
+
+  verticalAlign: theme.verticalAlign,
+  fontFamily: theme.fontFamily,
+  fontSize: theme.fontSize,
+
+  background: theme.background,
+  color: theme.color,
+  outline: theme.outline,
+
+  transition: theme.transition,
+
+  '&:hover': {
+    borderColor: theme['&:hover'].borderColor
+  },
+
+  '&:focus': {
+    outline: theme['&:hover'].outline,
+    outlineOffset: theme['&:hover'].outlineOffset
+  }
+});
 
 class Input extends React.Component {
   render() {
-    let { className, type, invalid, ...props } = this.props;
-
-    let _className = 'cf-input cf-input--' + type;
-
-    if (invalid) {
-      _className += ' cf-input--invalid';
-    }
-
-    if (className) {
-      _className += ' ' + className;
-    }
-
-    return <input className={_className} type={type} {...props} />;
+    const {
+      type,
+      name,
+      value,
+      onChange,
+      placeholder,
+      autoComplete,
+      invalid,
+      className
+    } = this.props;
+    return (
+      <input
+        type={type}
+        name={name}
+        value={value}
+        onChange={onChange}
+        placeholder={placeholder}
+        autoComplete={autoComplete}
+        invalid={invalid}
+        className={className}
+      />
+    );
   }
 }
 
@@ -35,4 +72,4 @@ Input.defaultProps = {
   type: 'text'
 };
 
-export default Input;
+export default createComponent(styles, Input);
--- a/packages/cf-component-input/src/InputTheme.js
+++ b/packages/cf-component-input/src/InputTheme.js
@@ -1,0 +1,27 @@
+export default baseTheme => ({
+  zIndex: 0,
+  width: '100%',
+  margin: '0 0 0.75rem',
+  padding: '0.45em 0.75em',
+  border: `1px solid ${baseTheme.color.hail}`,
+  borderRadius: baseTheme.borderRadius,
+
+  verticalAlign: 'middle',
+  fontFamily: baseTheme.fontFamily,
+  fontSize: '0.86667rem',
+
+  background: baseTheme.colorWhite,
+  color: baseTheme.color.charcoal,
+  outline: 'none',
+
+  transition: 'border-color 0.2s ease',
+
+  '&:hover': {
+    borderColor: '#256298'
+  },
+
+  '&:focus': {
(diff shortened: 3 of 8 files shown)


tajo:
Adding plugin `!important` should be in a separate PR. This is really hard to review.

koddsson:
Ah yeah, I should have put it in another commit.

koddsson:
Actually I did put it in a seperate PR but forgot to reset my `node_modules`. Doing that now :)

packages/cf-component-input/example/basic/component.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import Input from 'cf-component-input';
+import { Input } from 'cf-component-input';
```
hturan: Are we moving away from default exports?

packages/cf-component-input/src/Input.js
```diff
@@ -1,22 +1,37 @@
...
 
-class Input extends React.Component {
-  render() {
-    let { className, type, invalid, ...props } = this.props;
+const styles = ({ theme }) => ({
+  zIndex: theme.zIndex,
+  width: theme.width,
+  margin: theme.margin,
+  padding: theme.padding,
+  border: theme.border,
```
hturan: This might be missing the border-radius definition from our existing CSS.

packages/cf-component-input/example/basic/component.js
koddsson: [That is my understanding, yes :)](https://github.com/cloudflare/cf-ui/pull/285#discussion_r124845361)

packages/cf-component-input/src/Input.js
koddsson: I'll check that out, thanks.

koddsson: Nice catch 👌!

packages/cf-component-input/src/InputTheme.js
```diff
@@ -0,0 +1,27 @@
+export default baseTheme => ({
+  zIndex: 0,
+  width: '100%',
+  margin: '0 0 0.75rem',
+  padding: '0.45em 0.75em',
+  border: '1px solid #bdbdbd',
```
manatarms: This seems pretty common. If it's not present can we add it to style-const?

packages/cf-component-input/src/Input.js
```diff
@@ -1,22 +1,38 @@
...
-    return <input className={_className} type={type} {...props} />;
+  '&:focus': {
+    outline: theme['&:hover'].outline,
+    outlineOffset: theme['&:hover'].outlineOffset
+  }
+});
+
+class Input extends React.Component {
+  render() {
+    return <input {...this.props} />;
```
wyuenho: Do we still want to spread props?

koddsson: I'm not sure. I think I'd still want to, to be able to add things like `aria-*` classes and such? I remember needing it at a certain point but maybe that was just that `className` debate we had and doesn't apply anymore since we have fela?

wyuenho: @tajo What do you think? Should we just let React to warn us when an attribute unrecognized  by the DOM is passed in or should make some kind of a whitelist to filter invalid props for form elements? Do we just leave this as is?

tajo: Yea, we should avoid spreading and be explicit.

tajo: There are many many HTML attributes that you could pass down and maintaining (blacklists) filters is not a good approach.

koddsson: Could `fela` just read the proptypes of `<input />` (if any) so we don't need to manage this list by ourselves?

tajo: Actually you're right. 

> Could fela just read the proptypes of <input />

That's already happening.

So `createComponent()` should filter out all props that are not in propTypes. I think we are good. You should test it tho.

koddsson: Will do 👌

koddsson: >So `createComponent()` should filter out all props that are not in propTypes.

This doesn't seem to be happening.

<details>
    
   <summary>I wrote these tests</summary>

```jsx
  test('passes props from "PropTypes" down to wrapped component', () => {
    const Component = ({ foo }) => <div>{foo}</div>;
    Component.propTypes = {
      foo: PropTypes.string
    };
    const WrappedComponent = createComponent({}, Component);
    const instance = shallow(felaTestContext(<WrappedComponent foo="bar" />));
    expect(instance.find('Component').prop('foo')).toBe('bar');
  });

  test('does not pass down props that are not in "PropTypes" to wrapped component', () => {
    const Component = ({ foo }) => <div>{foo}</div>;
    Component.propTypes = {
      foo: PropTypes.string
    };
    const WrappedComponent = createComponent({}, Component);
    const instance = shallow(felaTestContext(<WrappedComponent bar="foo" />));
    expect(instance.find('Component').prop('bar')).toBe(undefined);
  });

  test('passes props from "native PropTypes" down to wrapped component', () => {
    const WrappedComponent = createComponent({}, <input />);
    const instance = shallow(
      felaTestContext(<WrappedComponent htmlFor="bar" type="number" />)
    );
    expect(instance.find('FelaComponent').prop('htmlFor')).toBe('bar');
    expect(instance.find('FelaComponent').prop('type')).toBe('number');
  });

  test('does not pass down props that are not in the "native PropTypes" to wrapped component', () => {
    const WrappedComponent = createComponent({}, <input />);
    const instance = shallow(felaTestContext(<WrappedComponent foo="bar" />));
    expect(instance.find('FelaComponent').prop('foo')).toBe(undefined);
  });
```

</details>

<details>

<summary>and got these failures</summary>

```
FAIL  packages/cf-style-container/test/index.js
  ● createComponent › does not pass down props that are not in "PropTypes" to wrapped component

    expect(received).toBe(expected)

    Expected value to be (using ===):
      undefined
    Received:
      "foo"

    Difference:

      Comparing two different types of values. Expected undefined but received string.

      at Object.<anonymous> (packages/cf-style-container/test/index.js:126:52)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
      at process._tickCallback (internal/process/next_tick.js:109:7)

  ● createComponent › does not pass down props that are not in the "native PropTypes" to wrapped component

    expect(received).toBe(expected)

    Expected value to be (using ===):
      undefined
    Received:
      "bar"

    Difference:

      Comparing two different types of values. Expected undefined but received string.

      at Object.<anonymous> (packages/cf-style-container/test/index.js:141:56)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
      at process._tickCallback (internal/process/next_tick.js:109:7)
```

</details>

koddsson: I'll whitelist the props and merge like that and we can revisit this later if we want.

Issue #210, opened 2017-05-26
Migrate `cf-component-dropdown` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

PR #291, opened 2017-07-03
feat(cf-component-dropdown): CSS to `fela`

koddsson:
BREAKING CHANGE: Major migration of styles from plain old CSS class
names to the CSS-in-JS framework `fela`.

Closes: #210

### DIFF
--- a/packages/cf-component-dropdown/src/Dropdown.js
+++ b/packages/cf-component-dropdown/src/Dropdown.js
@@ -1,7 +1,57 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import DropdownRegistry from './DropdownRegistry';
 import { canUseDOM } from 'exenv';
+import { createComponent } from 'cf-style-container';
+
+import DropdownRegistry from './DropdownRegistry';
+
+const styles = ({ theme, align }) => ({
+  position: 'absolute',
+  zIndex: 1,
+  minWidth: '10.66667rem',
+  margin: '0.5em 0 0',
+  padding: '0.33333rem 0',
+  listStyle: 'none',
+  background: theme.colorWhite,
+  border: `1px solid ${theme.colorGrayLight}`,
+  borderRadius: theme.borderRadius,
+  boxShadow: '0 3px 10px rgba(0, 0, 0, 0.2)',
+
+  left: align === 'left' ? 0 : 'initial',
+  right: align === 'right' ? 0 : 'initial',
+
+  animationName: {
+    '0%': {
+      display: 'none',
+      opacity: 0
+    },
+    '1%': {
+      display: 'block',
+      opacity: 0,
+      top: '80%'
+    },
+    '100%': {
+      display: 'none',
+      opacity: 1,
+      top: '102%'
+    }
+  },
+  animationDuration: '150ms',
+  animationTimingFunction: 'ease-out',
+
+  '&::before': {
+    content: "''",
+    display: 'block',
+    position: 'absolute',
+    bottom: '100%',
+    border: 'solid transparent',
+    borderWidth: '10px',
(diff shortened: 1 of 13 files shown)


packages/cf-component-dropdown/src/DropdownLink.js
```diff
@@ -1,8 +1,43 @@
...
 
+const styled = ({ theme }) => ({
+  position: 'relative',
+  zIndex: 1,
+
+  '&:first-child': {
+    borderTopWidth: 0
+  },
+
+  '& > a': {
```
koddsson: This will be fixed once #288 is merged

packages/cf-component-dropdown/test/Dropdown.js
```diff
@@ -43,51 +48,57 @@ describe('interactions', () => {
...
+      )
     );
-    const instance = render(dom, global.root);
+    const instance = dom.find('Dropdown').getNodes()[1];
+
+    expect(getFocusedHref()).toBe(null);
 
     instance.handleDocumentKeydown(down);
-    expect(getFocusedHref()).toBe('one');
+    setTimeout(() => expect(getFocusedHref()).toBe('one'), 0);
```
koddsson: I could try to do something here but am unsure what that would be. Some sort of callback comes to mind but I ended up doing this and seeing if someone _really_ hates it.

packages/cf-component-dropdown/test/Dropdown.js
```diff
@@ -1,25 +1,30 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
-import { trackFocus, getCurrentFocus } from '../../cf-test-focus/src/index';
```
koddsson: Maybe this can just be removed now? I didn't check if this was used in other places.

packages/cf-component-dropdown/src/Dropdown.js
```diff
@@ -81,4 +128,4 @@ Dropdown.childContextTypes = {
   dropdownRegistry: PropTypes.instanceOf(DropdownRegistry).isRequired
 };
 
-export default Dropdown;
+export default createComponent(styles, Dropdown, ['align']);
```
hturan: Why do we explicitly allow `align` here? Doesn't `createComponent` read from `propTypes`?

packages/cf-component-dropdown/src/DropdownLinkTheme.js
```diff
@@ -0,0 +1 @@
+export default baseTheme => ({});
```
hturan: What's the point in an empty theme? Do we need this?

packages/cf-component-dropdown/src/Dropdown.js
koddsson: Does it? I thought I always had to specify this 😱 or is that just when it's not part of the `propTypes`?

packages/cf-component-dropdown/src/DropdownLinkTheme.js
koddsson: I started doing empty themes after [@tajo told me that was ok](https://github.com/cloudflare/cf-ui/pull/281#pullrequestreview-46909882) 😛

packages/cf-component-dropdown/src/Dropdown.js
tajo: Yea, createComponent checks if you're wrapping a React component and [tries to add its propTypes](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-container/src/index.js#L14) into the third parameter.

koddsson: Ah ok , I'll remove this then.

Issue #220, opened 2017-05-26
Migrate `cf-component-card` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

koddsson:
Note that this includes abstracting the card title into a seperate component. See https://github.com/cloudflare/cf-ui/pull/188#pullrequestreview-40581836.

hturan:
> Note that this includes abstracting the card title into a seperate component. See #188 (review).

What about `footerMessage`? Same thing there? Are we planning on moving cf-ui from a mix of prop-driven composition models to a purely child-based JSX composition model?

koddsson:
@hturan I think so, yeah.

manatarms:
This seems done. Feel free to reopen if not.

PR #293, opened 2017-07-04
feat(cf-component-card): migrate to `fela`

koddsson:
BREAKING CHANGE: Deprecated `CardLoadingText` since it wasn't used at
all.

Additionally a major overhaul of a package internal warrants a major
version bump

Related-to: #220

### DIFF
--- a/packages/cf-component-card/src/CardBlock.js
+++ b/packages/cf-component-card/src/CardBlock.js
@@ -1,13 +1,20 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  padding: '1rem 1.4rem 1.26667rem',
+  borderTop: `1px solid ${theme.colorGrayLight}`,
+
+  '&:first-child': {
+    borderTop: 0
+  }
+});
 
 class CardBlock extends React.Component {
   render() {
-    const className = 'cf-card__block';
-
     return (
-      <div className={className}>
+      <div className={this.props.className}>
         {this.props.children}
       </div>
     );
@@ -18,4 +25,4 @@ CardBlock.propTypes = {
   children: PropTypes.node
 };
 
-export default CardBlock;
+export default createComponent(styles, CardBlock);
--- a/packages/cf-component-card/src/CardSection.js
+++ b/packages/cf-component-card/src/CardSection.js
@@ -1,18 +1,22 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
 
-class CardSection extends React.Component {
-  render() {
-    return (
-      <div
-        className={'cf-card__section cf-card__section--' + this.props.status}
-      >
-        {this.props.children}
-      </div>
-    );
+const styles = ({ theme, status }) => ({
+  display: 'table',
+  tableLayout: 'fixed',
+  width: '100%',
+  borderTop: `1px solid ${theme.colorGrayLight}`,
+  borderLeft: status === 'error' ? `3px solid ${theme.colorRed}` : 'initial',
+
+  '&:first-child': {
+    borderTop: 'initial'
   }
-}
+});
+
+const CardSection = ({ className, children }) => (
+  <div className={className}>{children}</div>
+);
 
 CardSection.propTypes = {
   status: PropTypes.oneOf(['default', 'error']),
@@ -23,4 +27,4 @@ CardSection.defaultProps = {
   status: 'default'
 };
 
-export default CardSection;
+export default createComponent(styles, CardSection, ['status']);
--- a/packages/cf-component-card/src/CardToolbarLink.js
+++ b/packages/cf-component-card/src/CardToolbarLink.js
@@ -1,6 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Link from 'cf-component-link';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ isActive }) => ({
+  position: 'relative',
+  display: 'inline-block',
+  padding: '0.73333rem 2em 0.73333rem 0.5rem',
+
+  '&::after': {
+    content: "'\\f003'",
+    fontFamily: '"cloudflare-font"',
+    fontStyle: 'normal',
+    fontWeight: 'normal',
+    marginTop: '-0.7em',
+
+    position: 'absolute',
+    top: '60%',
+    right: '1rem',
+    transition: 'transform 0.2s ease',
+    transform: isActive ? 'rotate(90deg)' : 'initial'
+  }
+});
 
 class CardToolbarLink extends React.Component {
   constructor(props) {
@@ -14,17 +35,11 @@ class CardToolbarLink extends React.Component {
   }
 
   render() {
-    let className = 'cf-card__toolbar_link';
-
-    if (this.props.isActive) {
-      className += ' cf-card__toolbar_link--open';
-    }
-
     return (
       <Link
         role="tab"
         id={this.props.id}
-        className={className}
+        className={this.props.className}
         onClick={this.handleClick}
       >
         {this.props.children}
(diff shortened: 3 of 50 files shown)


koddsson:
This is missing abstracting the title and footer into own components as described [here](https://github.com/cloudflare/cf-ui/pull/188#pullrequestreview-40581836).

packages/cf-component-card/src/CardToolbarLink.js
```diff
@@ -1,6 +1,27 @@
...
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ isActive }) => ({
+  position: 'relative',
+  display: 'inline-block',
+  padding: '0.73333rem 2em 0.73333rem 0.5rem',
+
+  '&::after': {
+    content: "'\\f003'",
+    fontFamily: '"cloudflare-font"',
```
tajo: We should probably use [this](https://github.com/cloudflare/cf-ui/pull/199/files#diff-642d1968224872e5886c42acd10d99b5R120) everywhere instead of relying on "outside" font face definition.

To make it easier, we could write a simple plugin that would replace:

```
createComponent({} => ({
  customFont: 'cloudflare'
}));
```

by 

```
createComponent({} => ({
  fontFace: {
    fontFamily: 'Cloudflare Icons',
    src: [`data:application/x-font-ttf;base64,AAAABBBBBCCCC...`]
  }
}));
``` 

Out of scope for this PR. Just an idea.

hturan:
This all seems fine. My only comment would be that having `Card`, `CardContent`, `CardSection` and `CardBlock` means that no-one will ever know what does what. How do I know which to use? What do they do? Can we get them more descriptive names?

packages/cf-component-card/src/CardSection.js
```diff
@@ -1,18 +1,22 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme, status }) => ({
+  display: 'table',
+  tableLayout: 'fixed',
+  width: '100%',
+  borderTop: `1px solid ${theme.colorGrayLight}`,
```
hturan: We currently only apply this rule on `& + &`. Does this affect anything?

koddsson:
@hturan I was thinking that the examples would clear most of that up but I agree. Only thing is that I'm missing the context in which these components were created so I'm not quite sure on other names.

Looking at https://cloudflare.github.io/cf-ui/#cf-component-card I feel like `CardSection` is something of a "row" and `CardControl` is something like a "column" but I'm not sure if those names would be any better 😱

packages/cf-component-card/src/CardToolbarLink.js
```diff
@@ -1,6 +1,27 @@
...
+    fontStyle: 'normal',
+    fontWeight: 'normal',
+    marginTop: '-0.7em',
+
+    position: 'absolute',
+    top: '60%',
```
hturan: This is currently `50%`.

packages/cf-component-card/src/CardSection.js
koddsson: I'm trying to remember back from migrating the `& + &`. If memory serves me right then it was causing the double border issue and this fixed it?

hturan:
How are we converting the `.cf-card .cf-table` and `.cf-card .cf-form` rules?

packages/cf-component-card/src/CardToolbarLink.js
koddsson: Yup! Changing it to `60%` fixed the alignment issues we currently have on production between the text and the arrow icon.

image

koddsson:
>How are we converting the .cf-card .cf-table and .cf-card .cf-form rules?

We're not. I feel like we had a conversation previously where I tested the table and form but I can't remember how that went now so I'll make sure to check those and come back with screenshots on how it looks.

packages/cf-component-card/src/CardBlock.js
```diff
@@ -1,13 +1,20 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  padding: '1rem 1.4rem 1.26667rem',
+  borderTop: `1px solid ${theme.colorGrayLight}`,
```
hturan: I can't find this in the existing CSS, only `.cf-card__block + .cf-card__block`.

packages/cf-component-card/src/CardToolbarLink.js
hturan: Awesome, thanks for this!

packages/cf-component-card/src/CardBlock.js
koddsson: Yeah but then I have the `&:first-child` selector below that makes sure that the border is `0` on the first child similar as the previous CSS rule.

packages/cf-component-card/src/CardToolbarLink.js
koddsson: No worries :)

packages/cf-component-card/src/CardBlock.js
hturan: But the previous CSS rule isn't about children, it's about siblings. `.cf-card__block + .cf-card__block` means _a `.cf-card__block` following a `.cf-card__block`_

hturan:
> We're not. I feel like we had a conversation previously where I tested the table and form but I can't remember how that went now so I'll make sure to check those and come back with screenshots on how it looks.

Appreciated, thanks!

packages/cf-component-card/src/CardBlock.js
koddsson: I thought I had checked this for sure 🤦‍♂️  I'll take another look.

koddsson: Ah! Now I remember something.

As far as I can remember I couldn't find any instances of where there are two `CardControl` components next to each other so I added it in the example. Since I'm not sure how it is supposed to look I took a stab at doing this by doing the above changes.

It seems to be working like expected even tho it's a child selector and not a sibling selector:

image

I feel like my understanding of `:first-child` is that it's not about children of the card block but the first child card block in a parent.

[According to this jsbin they are equivalent](http://jsbin.com/joxixosobo/edit?html,css,output)

With first child: 
image

Without first child:
image

koddsson:
>> We're not. I feel like we had a conversation previously where I tested the table and form but I can't remember how that went now so I'll make sure to check those and come back with screenshots on how it looks.

>Appreciated, thanks!

I'm looking at this now and I can't help but thinking that this should maybe be a integration in the dashboard rather than changes in the card component. Which would mean that I'd do those changes when I integrate the card into the dashboard after upgrading.

What do you think?

packages/cf-component-card/src/CardBlock.js
manatarms: @Neynah is working on multiple card content components. Maybe she can help out.

neynah: @koddsson I can say a lot of things about our card control patterns but I don't have the full picture of what's needed here so let me know if I/design team can be helpful in some way. :)

Some questions I'd like to ask anyway:
- Are the grey lines part of the pattern?
- Are there existing places in the UI where we have a block of text above a control?

koddsson: >* Are the grey lines part of the pattern?
Yup

>* Are there existing places in the UI where we have a block of text above a control?

Nope. There exists, in the CSS, some code to set the top border on all controls except the first one and I ported it directly here.

I'm a __huge__ fan of deleting unused code so if we can drop this, I am also down with that. In the past I haven't because people have blocked my PRs citing that we might need it in the future and I want to keep this CSS-in-JS migration going.

hturan:
> What do you think?

@koddsson You're totally right, sorry! That was a bit over-zealous there. This is an integration task rather than a component task 👍

koddsson:
No worries :)

Issue #216, opened 2017-05-26
Migrate `cf-component-callout` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

PR #294, opened 2017-07-05
feat(cf-component-callout): rewrite styles in fela

koddsson:
BREAKING CHANGE: Internals have been rewritten in a significant way.
This warrants a breaking change version bump.

Closes: #216


packages/cf-component-callout/src/Callout.js
```diff
@@ -1,16 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme, type = 'default' }) => {
+  return {
+    margin: '1rem',
```
manatarms: Base theme things?

PR #295, opened 2017-07-05
Migrate textarea & copyable-textarea to fela

koddsson:
Figured I'd do both at the same time.

### DIFF
--- a/packages/cf-component-textarea/src/TextareaTheme.js
+++ b/packages/cf-component-textarea/src/TextareaTheme.js
@@ -1,0 +1,32 @@
+export default baseTheme => ({
+  height: 'initial',
+  wordBreak: 'initial',
+  cursor: 'initial',
+  resize: 'initial',
+
+  zIndex: 0,
+  width: '100%',
+  margin: '0 0 0.75rem',
+  padding: '0.45em 0.75em',
+  border: `1px solid ${baseTheme.color.hail}`,
+  borderRadius: baseTheme.borderRadius,
+
+  verticalAlign: 'middle',
+  fontFamily: baseTheme.fontFamily,
+  fontSize: '0.86667rem',
+
+  background: baseTheme.colorWhite,
+  color: baseTheme.color.charcoal,
+  outline: 'none',
+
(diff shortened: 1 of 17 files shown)


packages/cf-component-textarea/src/TextareaTheme.js
```diff
@@ -0,0 +1,32 @@
...
+  height: 'initial',
+  wordBreak: 'initial',
+  cursor: 'initial',
+  resize: 'initial',
+
+  zIndex: 0,
+  width: '100%',
+  margin: '0 0 0.75rem',
+  padding: '0.45em 0.75em',
+  border: '1px solid #bdbdbd',
```
manatarms: Can we use the color from const? Or is this a completely different color?

koddsson: I'm not sure. In any case I can tell you that `#bdbdbd` is not a color we have in `cf-style-const`.

manatarms: @Neynah can this color be replaced or should we add it to style-const?

neynah: @koddsson Would love to see a screenshot to get more context. If it's really close to another color in our variables we should replace it. If it should be unique we can add a new variable for it. Right now it's hard for me to see where/how it's being used.

koddsson: I'm mostly just copy pasting this code over from our legacy CSS so I'm not really checking the values here but going over this I can see it's indistinguishable from they 'hail' color.

koddsson: Here's a comparison:

`#bdbdbd`
`#bcbec0` (hail)

I'll replace it with hail.

koddsson:
Hmmm this isn't failing locally, I'll have to take a look tomorrow.

koddsson:
If this next build fails I'll try to run the tests on a linux box since that's the only difference I can find between my development setup and CI.

koddsson:
Managed to make the build fail locally so I'll be working on that now!

PR #297, opened 2017-07-05
fix(saveFile): Use blob to trigger a file download to get around 64k …

wyuenho:
…attribute size limit

This approach works for IE11+

### DIFF
--- a/packages/cf-util-http-file/src/saveFile.js
+++ b/packages/cf-util-http-file/src/saveFile.js
@@ -1,4 +1,5 @@
 export default function saveFile(filename, file, mimeType) {
+  if (!filename || !file) throw new Error('filename and file are required');
   mimeType = mimeType || 'application/octet-stream';
   if ('msSaveBlob' in global.navigator) {
     global.navigator.msSaveBlob(new Blob([file]), filename);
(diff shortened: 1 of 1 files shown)


packages/cf-util-http-file/src/saveFile.js
```diff
@@ -1,22 +1,14 @@
...
   const link = global.document.createElement('a');
-  link.href = mimeType + ',' + global.encodeURI(file);
+  link.href = global.URL.createObjectURL(fileBlob);
   if (filename) link.download = filename;
 
-  // Dispatch fake click event to download the file
-  const event = global.document.createEvent('MouseEvents');
-  event.initEvent('click', true, true);
-  link.dispatchEvent(event);
+  link.click();
```
jrf0110: How sure are you that this works across all of our browsers? I noticed that MDN states `click()` will not initiate navigation on anchor tags https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click so it follows that it may not trigger a download.

wyuenho: Download is not considered navigation. I guess the better question is whether we should allow the file name be optional at all.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

wyuenho: Looks like is the head has a file name defined, it will be prioritized. I'll make the file name a requirement.

koddsson:
Ah if the commits had the package name in the optional scope (the part in the parenthesis) we could have published automatically.

PR #303, opened 2017-07-10
fix(cf-style-container): set themed component name

koddsson:
Sets the `displayName` property of themed components to
`Themed[Component name]` rather than the `ThemedComponent` it is set to
now.

This change will make it possible to select these components in
`enzyme` tests by their display names.

BREAKING CHANGE: The display name of themed components now have a
`Themed` prefix to their display names. So if you had a themed component
that has the name `Button` it will now be `ThemedButton` instead of
`ThemedComponent`.

Related-to: #300

### DIFF
--- a/packages/cf-style-container/src/index.js
+++ b/packages/cf-style-container/src/index.js
@@ -42,6 +42,7 @@ const applyTheme = (ComponentToWrap, ...themes) => {
     }
   }
 
+  ThemedComponent.displayName = `Themed${ComponentToWrap.displayName}`;
   ThemedComponent.childContextTypes = { theme: PropTypes.object };
   ThemedComponent.contextTypes = { theme: PropTypes.object };
   return ThemedComponent;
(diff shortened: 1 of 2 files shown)


packages/cf-style-container/src/index.js
```diff
@@ -42,6 +42,7 @@ const applyTheme = (ComponentToWrap, ...themes) => {
     }
   }
 
+  ThemedComponent.displayName = `Themed${ComponentToWrap}`;
```
hturan: Is `ComponentToWrap` a string? How does React coerce a component class to a string?

koddsson: WHOOPS. This was supposed to have a `.displayName`!

I went into the `node_modules` folder in the dashboard and edited the compiled source and then copy pasted it over to the `cf-ui` folder. Not sure how I missed this tho :E

koddsson: Very nice catch

koddsson: Imma add some tests so stuff like this is easier to catch and it's more apparent what we are doing here.

koddsson: Also coverage would be nice here.

hturan: Cool, no worries!

PR #304, opened 2017-07-11
fix(cf-component-form): FormFieldset invalid propTypes

jrf0110:
I saw that in our example, this component was failing validation since this prop was missing (and it's specified as required). However, it's not _really_ required. The component defaults to `vertical`. Well, that's not exactly true. All input that isn't `horizontal` is treated as `vertical`.

### DIFF
--- a/packages/cf-component-form/src/FormFieldset.js
+++ b/packages/cf-component-form/src/FormFieldset.js
@@ -60,10 +60,14 @@ class FormFieldset extends React.Component {
 }
 
 FormFieldset.propTypes = {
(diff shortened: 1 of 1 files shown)


tajo:
I am fine with removing `isRequired` but in that case I would add `FormFieldset.defaultProps`.

jrf0110:
> but in that case I would add `FormFieldset.defaultProps`.

Sure. Will do. It doesn't actually make any difference with the way it currently works, but if someone changes the implicit code later on, then having it explicitly default will be good.

koddsson:
Oh man that was supposed to be an approval! 😞

PR #306, opened 2017-07-12
fix(cf-component-input): spread input props

koddsson:
Instead of maintaning a 1-to-1 list of props that HTMLInputElement can accept we should just spread the ones we get from the parent on the `<input />` component.


hturan:
Thanks for this! 👌

koddsson:
This is stopping me from doing work so I'm just gonna merge with the one approval.

wyuenho:
^ poking for another approval isn't that hard, I was just going to approve it.

koddsson:
I know it's not hard and had done so with any reply so I just went for it.

PR #307, opened 2017-07-12
Allow passing selectorPrefix into cf-style-provider

sejoker:
Default value will be empty so no changes to the current behavior. We allow passing value through input parameter in case we need to use prefix (in case of class name collision)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "bootstrap": "lerna bootstrap",
     "build": "./scripts/build.sh",
     "lint": "eslint packages/*/src/**/*.js packages/*/example/**/*.js packages/*/test/**/*.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "bootstrap": "lerna bootstrap",
     "build": "./scripts/build.sh",
     "lint": "eslint packages/*/src/**/*.js packages/*/example/**/*.js packages/*/test/**/*.js",
-    "test": "cross-env BABEL_ENV=commonjs jest",
+    "test": "cross-env BABEL_ENV=commonjs jest --no-cache",
```
hturan: What's this change about?

sejoker: eliminates weird jest behavior when it's cache prevents tests from failing on source code changes.

PR #308, opened 2017-07-12
fix(cf-component-input): add blur/focus proptypes

koddsson:


packages/cf-component-input/src/Input.js
```diff
@@ -41,6 +41,8 @@ Input.propTypes = {
   name: PropTypes.string.isRequired,
   value: PropTypes.string.isRequired,
   onChange: PropTypes.func.isRequired,
+  onBlur: PropTypes.func,
+  onFocus: PropTypes.func,
```
manatarms: Should we also add an active state? I think our inputs do something on active.

PR #309, opened 2017-07-13
fix(cf-component-dropdown): text-align in theme

koddsson:
Add the `text-align` CSS property to the theme for the Dropdown component so that implementing parties can overwrite that rule.


koddsson:
Forgot to update the snapshots so this build is probably gonna fail D: I'll update them in a minute.

Issue #212, opened 2017-05-26
Migrate `cf-component-checkbox` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

toekneestuck:
Moved internal.

PR #310, opened 2017-07-13
feat(cf-component-checkbox): migrate css to fela

koddsson:
BREAKING CHANGE: Styles are now managed in the component using the CSS-in-JS framework `fela`. This internal re-write warrants a re-write.

Additionally the `onChange` callback on the input has changed. The callback now gives you the change event rather than just the `Event.EventTarget.checked` property.

Related-to: #212

### DIFF
--- a/packages/cf-component-checkbox/src/CheckboxGroup.js
+++ b/packages/cf-component-checkbox/src/CheckboxGroup.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Checkbox from './Checkbox';
 import includes from 'lodash/includes';
 
 class CheckboxGroup extends React.Component {
@@ -10,16 +9,16 @@ class CheckboxGroup extends React.Component {
   }
 
   handleChange(value, checked) {
-    const values = this.props.options
+    const values = this.props.children
       .filter(option => {
-        if (option.value === value) {
+        if (option.props.value === value) {
           return checked;
         }
 
-        return includes(this.props.values, option.value);
+        return includes(this.props.values, option.props.value);
       })
       .map(option => {
-        return option.value;
+        return option.props.value;
       });
 
     this.props.onChange(values);
@@ -27,19 +26,15 @@ class CheckboxGroup extends React.Component {
 
   render() {
     return (
-      <div className="cf-checkbox__group">
-        {this.props.options.map(option => {
-          return (
-            <Checkbox
-              key={option.name}
-              label={option.label}
-              name={option.name}
-              value={option.value}
-              checked={includes(this.props.values, option.value)}
-              onChange={e => this.handleChange(option.value, e.target.checked)}
-            />
-          );
-        })}
+      <div>
+        {React.Children.map(this.props.children, Checkbox =>
+          React.cloneElement(Checkbox, {
+            key: Checkbox.props.name,
+            checked: includes(this.props.values, Checkbox.props.value),
+            onChange: event =>
+              this.handleChange(Checkbox.props.value, event.target.checked)
+          })
+        )}
       </div>
     );
   }
--- a/packages/cf-component-checkbox/example/basic/component.js
+++ b/packages/cf-component-checkbox/example/basic/component.js
@@ -1,5 +1,23 @@
 import React from 'react';
-import { Checkbox, CheckboxGroup } from 'cf-component-checkbox';
+import {
+  Checkbox,
+  CheckboxUnstyled,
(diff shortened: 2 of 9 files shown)


tajo:
> Additionally the onChange callback on the input has changed. The callback now gives you the change event rather than just the Event.EventTarget.checked property.

Why? It seems like an implementation detail that doesn't need to be exposed.

packages/cf-component-checkbox/src/CheckboxGroup.js
```diff
@@ -1,8 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Checkbox from './Checkbox';
+import { applyTheme } from 'cf-style-container';
 import includes from 'lodash/includes';
 
+import CheckboxUnstyled from './Checkbox';
+import CheckboxTheme from './CheckboxTheme';
+
+const Checkbox = applyTheme(CheckboxUnstyled, CheckboxTheme, theme => ({
```
tajo: How about adding a bool prop `group` to the `cf-component-checkbox/checkbox`? Then you can do `marginTop: group ? ...` there. 

It's less code and you can reuse the original Checkbox component instead of creating this new one.

packages/cf-component-checkbox/src/CheckboxGroup.js
```diff
@@ -27,7 +38,7 @@ class CheckboxGroup extends React.Component {
 
   render() {
     return (
-      <div className="cf-checkbox__group">
+      <div>
         {this.props.options.map(option => {
```
tajo: I would probably prefer this API (and utilize React.Children.map +  React.cloneElement)

```jsx
<CheckboxGroup>
  <Checkbox />
  <Checkbox />
</CheckboxGroup>
```

instead of 

```jsx
<CheckboxGroup options={...} />
```

so it follows the `ButtonGroup` and `Button` example. The later seems more as a builder pattern (more restrictive). However, since you are just keeping the existing solution here, let's revisit this another day.

koddsson:
>> Additionally the onChange callback on the input has changed. The callback now gives you the change event rather than just the Event.EventTarget.checked property.

>Why? It seems like an implementation detail that doesn't need to be exposed.

It matches the behaviour of `cf-component-input` as well as giving you access to the whole event in case you'd like to use the target element or any other data in the event.

packages/cf-component-checkbox/src/CheckboxGroup.js
koddsson: Sounds good to me.

packages/cf-component-checkbox/src/CheckboxGroup.js
koddsson: Yeah I think I'd like to try to get stuff to `fela` first and then look at stuff like this. Maybe we could have a codebase review or something to look over the codebase and clean some of the stuff like this up.

packages/cf-component-checkbox/src/CheckboxGroup.js
tajo: @koddsson bump :innocent:

koddsson: Forgot about this. I'll update it.

koddsson: Actually, I'm thinking about this and is this the pattern we want to do? Won't we just end up with a lot of props for every possible integration situation and a lot of complexity in the `Checkbox` component?

tajo: Hm, are there any other props that you would need to add? Giving the component some contextual awareness through an extra prop (in other words telling it "you are in a group") seems reasonable to me. It worked out pretty well for the Button.

koddsson: There's nothing I'd want to add right now but if we follow this path I can see us having: `inAbox`, `inACardSection`, `inACardBlock`, `inAPopUpModal`, `inInput`, etc. rather than putting the styles that are needed for each implementation by using `applyTheme` to override the styles. The latter seems more reasonable to me and it's kind of the path I've been going down with a lot of the components I've migrated.

Maybe I should just pull the `<Checkbox />` components out like I did in the radio migration (#333) and you suggested below. It makes a lot of these issues disappear.

koddsson: I'll rip it out and see how that goes.

koddsson:
What do you think about this @tajo ?

PR #313, opened 2017-07-18
feat(cf-component-input): allow text-indent on input

koddsson:
In fela we need to add all properties we'd like to override with other values into the component.

This PR adds the `text-indent` key to the style object of the Input component.


tajo:
The PR is fine but the description is not right?

> This PR adds the text-indent property to the prop types of the Input component.

Not really? I see it's only adding a new CSS prop textIndent.

koddsson:
lol yeah, I'll update the commit message.

Issue #222, opened 2017-05-26
Migrate `cf-component-radio` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

koddsson:
Installed but not implemented in the dashboard.

PR #316, opened 2017-07-18
feat(cf-component-radio): rewrite styling in fela

koddsson:
BREAKING CHANGE: This change is a significant change in the component
internals warranting a major bump in version number.

Closes: #222

### DIFF
--- a/packages/cf-component-radio/src/Radio.js
+++ b/packages/cf-component-radio/src/Radio.js
@@ -21,14 +21,14 @@ const styles = ({ theme }) => ({
 const Input = createComponent(
   ({ theme, checked }) => ({
     zIndex: 0,
-    border: '1px solid #bdbdbd',
+    border: `1px solid ${theme.color.hail}`,
 
     verticalAlign: 'middle',
     fontFamily: 'Open Sans,Helvetica,Arial,sans-serif',
     fontSize: '0.86667rem',
 
     background: theme.colorWhite,
-    color: '#333333',
+    color: theme.color.charcoal,
     outline: 'none',
 
     transition: 'border-color 0.2s ease',
@@ -48,7 +48,7 @@ const Input = createComponent(
     },
 
     '&:focus': {
-      borderColor: '#2f7bbf',
+      borderColor: theme.color.marine,
       outline: '5px auto -webkit-focus-ring-color',
       outlineOffset: -1
     },
--- a/packages/cf-component-radio/src/Radio.js
+++ b/packages/cf-component-radio/src/Radio.js
@@ -1,6 +1,86 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  cursor: theme.cursor,
+  display: theme.display,
+  minHeight: theme.minHeight,
+  paddingLeft: theme.paddingLeft,
+  marginTop: theme.marginTop,
+
+  '&:first-child': {
+    marginTop: theme['&:first-child'].marginTop
+  },
+
+  '&:hover input': {
+    borderColor: theme['&:hover input'].borderColor
+  }
+});
+
+const Input = createComponent(
+  ({ theme, checked }) => ({
+    zIndex: 0,
+    border: '1px solid #bdbdbd',
+
+    verticalAlign: 'middle',
+    fontFamily: 'Open Sans,Helvetica,Arial,sans-serif',
+    fontSize: '0.86667rem',
+
+    background: theme.colorWhite,
+    color: '#333333',
+    outline: 'none',
+
+    transition: 'border-color 0.2s ease',
+
+    position: 'relative',
+    height: 15,
+    width: 15,
+    margin: '1px 0 0',
+    padding: 0,
+    lineHeight: 'normal',
+    appearance: 'none',
+    top: -1,
+    borderRadius: '50%',
+
+    '&:hover': {
+      borderColor: '#256298'
+    },
+
+    '&:focus': {
+      borderColor: '#2f7bbf',
+      outline: '5px auto -webkit-focus-ring-color',
+      outlineOffset: -1
+    },
+
+    '&::before': {
+      content: "''",
+      position: 'absolute',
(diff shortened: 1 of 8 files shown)


koddsson:
I'll go over the colors and try to find matches to the ones I have.

packages/cf-component-radio/src/Radio.js
```diff
@@ -1,6 +1,86 @@
...
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  cursor: theme.cursor,
+  display: theme.display,
+  minHeight: theme.minHeight,
+  paddingLeft: theme.paddingLeft,
+  marginTop: theme.marginTop,
+
+  '&:first-child': {
```
tajo: This is always going to match `<Input />`. Can you just add a prop `isFirst` to the `<Input />` instead of this first-child selector?

packages/cf-component-radio/src/RadioGroup.js
```diff
@@ -1,11 +1,28 @@
...
-import Radio from './Radio';
+import { createComponent, applyTheme } from 'cf-style-container';
+
+import RadioUnstyled from './Radio';
+import RadioTheme from './RadioTheme';
+
+const Radio = applyTheme(RadioUnstyled, RadioTheme, () => ({
+  marginTop: '1em',
+
+  '&:first-child': {
```
tajo: same thing, could we replace it by a prop?

packages/cf-component-radio/src/Radio.js
koddsson: Maybe, this is all copy-pasted and I'm not actually writing this code with any active thoughts other than trying to get the visuals to match. I can take a look at this but am unsure to what the problem is that I'm supposed to solve. Maybe explaining that would help me do the suggested change :)

tajo: **This works.**, However, it breaks one of the [CSS in JS principles - isolation](https://speakerdeck.com/vjeux/react-css-in-js). It's actually nicely described there. 

So what's happening? Component `Radio` is changing styles of component `Input` through a CSS first-child selector.  Component Input has no clue that this is happening. If you are an author of Input, you are losing ability to introduce safe upgrades since people "go around your API". API for our components are props. Does it make sense?

Anyway, if you don't want to thing about this now, it's ok to leave it as it is. We can revisit later.

koddsson: >Anyway, if you don't want to think about this now, it's ok to leave it as it is. We can revisit later.

Sounds goooooooooooooood

koddsson:
>@koddsson Did you get a chance to work with design and consolidate colors into cf-style-const?

Nah I forgot! I'll take a look at the colors and see if I can find matches right now.

packages/cf-component-radio/src/RadioTheme.js
```diff
@@ -0,0 +1,15 @@
...
+  minHeight: '1em',
+  paddingLeft: '2em',
+  marginTop: 'initial',
+
+  '&:first-child': {
+    marginTop: 'initial'
+  },
+
+  '&:hover input': {
+    borderColor: '#256298'
```
koddsson: This color: `#256298`.

Marine:       `#2F7BBF`.
Rain:            `#408BC9`.

Is either one of those close enough to pick?

packages/cf-component-radio/src/Radio.js
```diff
@@ -1,6 +1,86 @@
...
+
+  '&:hover input': {
+    borderColor: theme['&:hover input'].borderColor
+  }
+});
+
+const Input = createComponent(
+  ({ theme, checked }) => ({
+    zIndex: 0,
+    border: '1px solid #bdbdbd',
```
koddsson: This color: `#bdbdbd`.

storm:        `#808285`
hail:            `#BCBEC0`
smoke:        `#e0e0e0`

Hail feels like the closest one, I'll go with that.

packages/cf-component-radio/src/Radio.js
```diff
@@ -1,6 +1,86 @@
...
+
+    verticalAlign: 'middle',
+    fontFamily: 'Open Sans,Helvetica,Arial,sans-serif',
+    fontSize: '0.86667rem',
+
+    background: theme.colorWhite,
+    color: '#333333',
```
koddsson: Charcoal.

packages/cf-component-radio/src/Radio.js
```diff
@@ -1,6 +1,86 @@
...
+    appearance: 'none',
+    top: -1,
+    borderRadius: '50%',
+
+    '&:hover': {
+      borderColor: '#256298'
+    },
+
+    '&:focus': {
+      borderColor: '#2f7bbf',
```
koddsson: Marine.

packages/cf-component-radio/src/RadioTheme.js
tajo: Deepsea: #1E4E79

tajo: Deepsea: `#1E4E79`

packages/cf-component-radio/src/Radio.js
neynah: 👍

packages/cf-component-radio/src/Radio.js
neynah: 👍

packages/cf-component-radio/src/Radio.js
neynah: 👍 (Annoyingly approving stuff because @tajo asked me to)

tajo: That's what I got for trying to keep people in the loop. :grimacing:

tajo: Ironically enough your nickname consists of `nay` and `nah`.

packages/cf-component-radio/src/RadioTheme.js
neynah: If this color is only used for borders then it's fine to use Deepsea. If it's used for hovers as well then the colors (`#256298` and `#1E4E79` are probably too far apart (and should be added to the palette).

PR #317, opened 2017-07-18
chore(cf-style-const): rename gradient keys in theme

manatarms:
BREAKING CHANGE: This change will break all places that use gradients from theme


manatarms:
Places gradients are used right now.

In www-next
1) overview page - CardHeader

In cf-ux
1) OnboardingContentView
2) OnboardingView
3) cf-component-presentation -> TopGradient

packages/cf-builder-pagination/test/__snapshots__/PaginationBuilder.js.snap
```diff
@@ -26,7 +26,7 @@ exports[`should render pagination 1`] = `
       </a>
     </li>
     <li
-      className="m p ad ae   s af ag"
```
manatarms: Fela changed the spaces on the snapshots for some reason. Not sure why.

koddsson:
Did publishing with the `chore` keyword work alright?

manatarms:
@koddsson It was beautiful. It just figured out everything that needed a bump, which in this case was `cf-style-const`. So nice! Thanks for this.

PR #314, opened 2017-07-18
Migate tabs to fela

koddsson:
I added a second commit that fixes this gap in the examples:

Before:

image

After:

image

PR #318, opened 2017-07-19
fix(cf-component-tabs): set the correct colors

koddsson:
In #314 we errounisly made assumptions where colors are kept in the
theme. Trying to access the white color at `theme.color.white` and the
smoke color at `theme.color.smoke` resulted in `undefined` values that
`fela` swallows up and not emitting a rule.

This patch references the colors in the correct location as well as
minor housekeeping to keep the tabs theme a bit tidier.

### DIFF
--- a/packages/cf-component-tabs/src/Tabs.js
+++ b/packages/cf-component-tabs/src/Tabs.js
@@ -34,50 +34,50 @@ const TabsGroup = createComponent(
 
 const TabItem = createComponent(
   ({ theme, selected }) => ({
-    flex: theme.flex,
-    background: theme.background,
-    color: theme.color,
-    cursor: theme.cursor,
-    display: theme.display,
-    margin: theme.margin,
-    padding: theme.padding,
-    position: theme.position,
-    textAlign: theme.textAlign,
-    verticalAlign: theme.verticalAlign,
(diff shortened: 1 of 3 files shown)


packages/cf-component-tabs/src/Tabs.js
```diff
@@ -34,50 +34,50 @@ const TabsGroup = createComponent(
...
-    display: theme.display,
-    margin: theme.margin,
-    padding: theme.padding,
-    position: theme.position,
-    textAlign: theme.textAlign,
-    verticalAlign: theme.verticalAlign,
-    outline: theme.outline,
-    userSelect: theme.userSelect,
+    flex: theme.item.flex,
+    background: theme.item.background,
```
manatarms: Just wondering if you know why it did this. Seems interesting. Does it do it to all our components?

tajo:
So why did you move everything into `theme.item`? To get an error without the theme?

koddsson:
I moved it so it wouldn't clash with the styles being set on the "root" element.

koddsson:
This passes locally :(((((((((((((((

koddsson:
Not sure what was wrong with my local environment but deleting everything was the answer here.

Issue #231, opened 2017-05-26
Migrate `cf-component-list` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

koddsson:
Setting low priority since this isn't being used in the dashboard.

PR #319, opened 2017-07-19
feat(cf-component-list): rewrite styling in fela

koddsson:
BREAKING CHANGE: This change is a significant change in the component
internals warranting a major bump in version number.

Closes: #231


manatarms:
You need to update the snapshots after the fela space issue.

packages/cf-component-list/src/ListItem.js
```diff
@@ -1,19 +1,15 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = () => ({});
```
tajo: no styles?

koddsson: Nope, there was nothing in the dashboard for `ListItem`

PR #321, opened 2017-07-24
feat(cf-component-tabs): `TabsItem` own component

koddsson:
In order to be able to apply styles to a "tab item" we need to abstract
it out into it's own component. That can be optionally be passed down to
the `Tabs` component defaulting to the styles previously applied.

This change is backwards compatible. See
`packages/cf-component-tabs/example/basic/component.js` for the optional
component property.

### DIFF
--- a/packages/cf-component-tabs/example/basic/component.js
+++ b/packages/cf-component-tabs/example/basic/component.js
@@ -1,5 +1,15 @@
 import React from 'react';
-import { Tabs, TabsPanel } from 'cf-component-tabs';
+import {
+  Tabs,
+  TabsItemUnstyled,
+  TabsItemTheme,
+  TabsPanel
+} from 'cf-component-tabs';
+import { applyTheme } from 'cf-style-container';
+
+const Tab = applyTheme(TabsItemUnstyled, TabsItemTheme, () => ({
+  background: '#FFFFF0'
+}));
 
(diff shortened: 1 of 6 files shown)


packages/cf-component-tabs/example/basic/component.js
```diff
@@ -1,5 +1,15 @@
...
+import {
+  Tabs,
+  TabsItemUnstyled,
+  TabsItemTheme,
+  TabsPanel
+} from 'cf-component-tabs';
+import { applyTheme } from 'cf-style-container';
+
+const Tab = applyTheme(TabsItemUnstyled, TabsItemTheme, () => ({
+  background: '#FFFFF0'
```
manatarms: I feel like I say this too much but style-const?. Sorry :(
@neynah can help.

neynah: Where is this yellowy-white color being used exactly? Seems like it could actually be white... but a screenshot/additional context would help. :)

koddsson: This is not in the component at all but rather just in the example to showcase how you can override the styles of a `TabItem` component.

manatarms: Lol. Didn't see the file path. LGTM

PR #322, opened 2017-07-24
Test clean up

koddsson:
Cleaning up all console logs and add tests to disallow as well as trying to run tests faster on CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ before_script:
 script:
   - yarn run build
   - yarn run lint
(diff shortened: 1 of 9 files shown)


koddsson:
Here's the difference from those two commits on Travis:
image

koddsson:
`cf-component-dropdown` and `cf-style-container` are still slow since they need to be mounted and that's more work then I wanna do right now.

koddsson:
Lean, clean tests are pretty nice 💨

Issue #227, opened 2017-05-26
Migrate `cf-component-dynamic-content` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

koddsson:
There don't seem to be any CSS classes associated with this component. The class name is just used to target other components.

PR #323, opened 2017-07-25
feat(cf-component-dynamic-content): css2fela

koddsson:
Convert the `cf-component-dynamic-content` to fela.

Related-to: #227

### DIFF
--- a/packages/cf-component-dynamic-content/src/DynamicContent.js
+++ b/packages/cf-component-dynamic-content/src/DynamicContent.js
@@ -1,12 +1,16 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  display: theme.display
+});
 
 class DynamicContent extends React.Component {
   render() {
     return (
       <div
-        className="cf-dynamic-content"
+        className={this.props.className}
         dangerouslySetInnerHTML={this.props.dangerouslySetInnerHTML}
       />
     );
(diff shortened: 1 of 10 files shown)


test_setup/patch-console.js
```diff
@@ -10,6 +10,8 @@ global.console.trace = () => {
   throw new Error("Please don't call console.trace in the codebase");
 };
 
-global.console.log = () => {
+const log = global.console.log;
+global.console.log = message => {
+  log(message);
```
koddsson: I added this here so I can add console logs temporarily when debugging but still have this fail.

packages/cf-component-dynamic-content/src/DynamicContent.js
```diff
@@ -1,12 +1,16 @@
 import React from 'react';
-
 import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  display: theme.display
```
hturan: It looks like we override `color` and `textDecoration` internally. Will we still be able to do so here?

packages/cf-component-dynamic-content/src/DynamicContentTheme.js
```diff
@@ -0,0 +1,3 @@
+export default baseTheme => ({
+  display: 'initial'
```
sejoker: `initial` is the default  value, safe to remove

packages/cf-component-dynamic-content/src/DynamicContent.js
koddsson: We talked about this IRL

packages/cf-component-dynamic-content/src/DynamicContentTheme.js
koddsson: I needed here if I want to override it.

PR #325, opened 2017-07-25
fix(cf-component-input): Input[type] missing valid options

jrf0110:


koddsson:
The way we have `cf-ui` set up right now means that we need to re-implement all these browser APIs and keep them in sync which seems weird to me. Not a comment on this PR but rather this whole system.

manatarms:
Can we close this if everyone agrees?

PR #326, opened 2017-07-26
feat(cf-component-tabs): allow border override

koddsson:


sejoker:
what is the rational to make the change, where the border is needed to be overridden?

koddsson:
In the DNS card.

koddsson:
There's a tab component within a tab component there.

manatarms:
> There's a tab component within a tab component there.
LOL

PR #327, opened 2017-07-27
chore(cf-component-box): Added missing display table props

manatarms:


koddsson:
** Not a comment on this PR but rather cf-ui as a whole **

It's kind of ridiculous that we need to do stuff like this IMHO.

manatarms:
@koddsson I totally agree. But I feel this is a side effect of our Fela migration. Once we figure all the props we need, it should stabilize :)

koddsson:
I'm not sure @manatarms, but I guess time will tell :)

tajo:
> It's kind of ridiculous that we need to do stuff like this IMHO.

You don't *really* need to do that. You can use `createComponent`. `<Box />` was never meant as a full-featured replacement. It's more an experiment and I think it's fine for extra margin/padding but we definitely don't have to copy&paste the whole CSS spec into it.

koddsson:
Does it just work like this?

```js
createComponent({ backgroundColor: 'red' }, Box)
```

PR #330, opened 2017-07-28
Deprecate packages

koddsson:
If this PR gets merged I'll manually deprecate the packages on npm.


manatarms:
There are quite a few instances of `<Flex/>` and `<Loading/>` in the codebase. How do we plan on supporting these in the future?

tajo:
I'm ok with removing layout, flex and viewport.

I would keep the loading around. We are planning to start using react-loadable all over the place, so I would like to use this package for an universal "component loading state".

Also I see a lot of value in typography. Maybe it just needs an update? What does @mrmrs think?

koddsson:
>There are quite a few instances of <Flex/> and <Loading/> in the codebase. How do we plan on supporting these in the future?

Those will stay in the codebase for now. Deprecating them (as I plan to do it) will mean that the code won't be in cf-ui, the packages will receive no further updates and npm/yarn will display a deprecation warning when installing.

>I would keep the loading around. We are planning to start using react-loadable all over the place, so I would like to use this package for a universal "component loading state".

Why not just use the icon the `Loading` component wraps around directly?

>Also I see a lot of value in typography.

I don't. If for not any other reason than it's not used at all in the dashboard with (as far as I know) any plans to use them in the future. I say we deprecate and then we can always go digging for the code we want if we end up wanting any of it.

I just _really_ don't want to put in all the work to migrate something on the off chance we _might_ use it in the future.

sejoker:
I don't agree to deprecate `cf-component-loading`. It's encapsulates how we implement `loading` signifier to the user.
>Why not just use the icon the Loading component wraps around directly?
again, encapsulation, if we will change internal implementation in the future we won't need to update codebase in all the places where it's used.

sejoker:
fine to deprecate rest of them.

koddsson:
>Again, encapsulation, if we will change internal implementation in the future we won't need to update codebase in all the places where it's used.

Sounds good.

I think there's a line that components can fall on either side of where components benefit of encapsulation or the cost of maintaining the encapsulation becomes a pain. `Loading` probably falls within the former group.

jwineman:
There might be a11y implications to keeping `<Loading/>` separate too.

koddsson:
Rebased the loading commit out.

tajo:
Ok, I'm fine with this.

koddsson:
```sh
☁  cf-ui [master] npm deprecate cf-component-typography "Please consider alternatives to cf-component-typography" 
☁  cf-ui [master] npm deprecate cf-component-layout "Please consider alternatives to cf-component-layout"    
☁  cf-ui [master] npm deprecate cf-component-layout "Please consider alternatives to cf-component-flex"  
☁  cf-ui [master] npm deprecate cf-component-layout "Please consider alternatives to cf-component-viewport"
```

PR #331, opened 2017-07-28
feat(cf-style-const): Add a new color

tajo:
Approved by Nena.


manatarms:
Will you be updating the styleguide?

neynah:
Yay 🎈

Issue #213, opened 2017-05-26
Migrate `cf-component-select` to fela.


koddsson:
Migrate the component to the CSS-in-JS framework `fela`

wyuenho:
This one is going to be hard

koddsson:
Whoops!

PR #335, opened 2017-08-01
feat(cf-component-select): migrate select to fela

koddsson:
Migrate the `cf-component-select` component to use the CSS-in-JS
framework fela.

Closes: #213

### DIFF
--- a/packages/cf-component-select/src/Select.js
+++ b/packages/cf-component-select/src/Select.js
@@ -1,10 +1,39 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactSelect from 'react-select';
+import { createComponent } from 'cf-style-container';
+
+const styles = ({ theme }) => ({
+  flex: theme.flex,
+  margin: theme.margin,
+  width: theme.width,
+  display: theme.display,
+  verticalAlign: theme.verticalAlign
+});
+
+const Label = createComponent(
+  ({ theme }) => ({
+    display: 'block',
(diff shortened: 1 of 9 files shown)


packages/cf-component-select/src/Select.js
```diff
@@ -1,10 +1,39 @@
...
+  width: theme.width,
+  display: theme.display,
+  verticalAlign: theme.verticalAlign
+});
+
+const Label = createComponent(
+  () => ({
+    display: 'block',
+    fontSize: '0.7em',
+    color: '$cf-select-label-color',
```
tajo: ¯\_(ツ)_/¯

koddsson: lol

packages/cf-component-select/test/Select.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import toJson from 'enzyme-to-json';
-import Select from '../../cf-component-select/src/index';
+import { Select } from '../../cf-component-select/src/index';
 
 test('should render', () => {
   const tree = shallow(<Select />);
```
tajo: `mount()` would be better. 

1. the snapshots contains only the Fela wrapper
2. `shallow()` doesn't pass context (theme) down correctly

koddsson: Sure

PR #341, opened 2017-08-04
chore(deployment): Automatically publish packages

koddsson:
This patch will instruct the CI system to run a deploy command (yarn run publish -- --yes) when a commit lands on the `master` branch.

This will mean that developers won't have to run these commands from local machines to publish packages anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,8 @@ script:
   - yarn run build
   - yarn run lint
   - yarn test -- --runInBand
+deploy:
+  provider: script
+  script: yarn run publish -- --yes
+  on:
+    branch: production
--- a/README.md
+++ b/README.md
@@ -55,6 +55,24 @@ open index.html
 
 To get started contributing please see [CONTRIBUTING.md](CONTRIBUTING.md)
 
+## Deploying new packages
+
+When a commit hits the `master` branch it will go through the CI pipeline which
+will at the end publish the packages to the npm registry given that the commits
+have the correct conventional commits format.
+
+Here's a example conventional commit message for reference. Note that the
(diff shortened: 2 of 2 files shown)


koddsson:
Let's see if this thing will actually work :D

jwineman:
image

manatarms:
So question. What about when we publish our docs directly to master? I usually have (publish docs) as the `<type>`. Should I now just say (publish) instead.

.travis.yml
```diff
@@ -7,3 +7,8 @@ script:
   - yarn run build
   - yarn run lint
   - yarn test -- --runInBand
+deploy:
+  provider: script
+  script: yarn run publish -- --yes
+  on:
+    branch: production
```
koddsson: WHOOPS

koddsson: This was supposed to be `master` :D

koddsson:
>I usually have (publish docs) as the <type>. Should I now just say (publish) instead.

I'm not sure I follow, where and to whom are you saying (publish)?

manatarms:
Lol sorry. Should have been more clear. So when I push to master to publish styleguide docs, I usually have my commit message say `chore(some-scope): Publish docs`. So when I push now to master. Do I need to also say `chore(some-scope): Publish docs + deployment` because CI is technically also publishing packages?

koddsson:
Nope, no need to put any indicator so that Travis will publish. It will try to publish any commit that lands on the `master` branch.

tajo:
@manatarms Since it doesn't change anything in `/packages` and I assume that lerna just exits the publishing (no diff, nothing to publish).

PR #342, opened 2017-08-04
chore(deployment): publish from the master branch

koddsson:
In 264539a we set up some pretty sweet continues deployment system but erroneously set the deployment branch to `production` when our "production" branch is in fact called "master".

This patch changes the Travis "deploy" branch to `master.


koddsson:
image

PR #343, opened 2017-08-04
chore(deployment): install deps before deploying

koddsson:
Apparently, we need to install all dependencies before deploying since the deployment steps run in a different context than the previous build and tests contexts.


koddsson:
I KEEP PRESSING BUTTONS AND NOTHING HAPPENS!

koddsson:
Third time's the charm?

PR #344, opened 2017-08-07
New consts & removal of flex

tajo:
- new color used by navigation
- new special media query used by navigation
- removal of flex from the form component, it's not needed and it breaks cf-styleguide


packages/cf-style-provider/src/createRenderer.js
```diff
@@ -19,7 +19,8 @@ const mediaQueries = {
   mobileWide: `@media (min-width: ${variables.breakpoints.mobileWide})`,
   tablet: `@media (min-width: ${variables.breakpoints.tablet})`,
   desktop: `@media (min-width: ${variables.breakpoints.desktop})`,
-  desktopLarge: `@media (min-width: ${variables.breakpoints.desktopLarge})`
+  desktopLarge: `@media (min-width: ${variables.breakpoints.desktopLarge})`,
+  hover: `@media (hover)`
```
koddsson: Are we relying on this somewhere or is this just a enhancement? It doesn't seem to [be supported in Firefox](https://caniuse.com/#feat=css-media-interaction).

packages/cf-style-provider/src/createRenderer.js
```diff
@@ -47,6 +48,7 @@ const createRenderer = opts => {
     enhancers,
     selectorPrefix: usedOpts.selectorPrefix,
     mediaQueryOrder: [
```
koddsson: This list seems weird and redundant. Can't this just be `Object.values(mediaQueries).map(removePrefix)`?

packages/cf-style-provider/src/createRenderer.js
tajo: Yep, I use it strictly for enhancements.

packages/cf-style-provider/src/createRenderer.js
tajo: It theory. But objects don't have to preserve the order of properties so it's not safe. That's why that API is using an array in the first place.

PR #345, opened 2017-08-07
Added 'vertical' prop to ButtonGroup

hturan:
This PR adds a `vertical` prop to `ButtonGroup`, allowing for vertical groups of buttons, as seen in the below screenshot:

screen shot 2017-08-07 at 11 42 11

### DIFF
--- a/packages/cf-component-button/src/Button.js
+++ b/packages/cf-component-button/src/Button.js
@@ -53,14 +53,24 @@ const before = (fadeZoomIn, loading) => {
   };
 };
 
-const marginLeft = (marginLeft, group, spaced) => {
-  if (!group) {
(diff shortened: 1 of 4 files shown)


koddsson:
Wait with merging this until I figure the deployment out? It shouldn't be more than an hour 🙏

hturan:
> Wait with merging this until I figure the deployment out? It shouldn't be more than an hour 🙏

Sure thing!

koddsson:
I rolled back the automatic deployment changes. Feel free to publish manually when ever you'd lik.e

hturan:
> I like the feature! One thing is we could do better IMO if instead adding ad hoc vertical property we follow flex with direction value of column and row.

Good idea! All done 👍

PR #354, opened 2017-08-22
feat(cf-style-const): add 'cherry' color

hturan:
This adds a new color to `cf-style-const` named 'cherry'.


manatarms:
Will you update the styleguide too? I can do it if you have too much stuff.

hturan:
@manatarms I've never done it before, what does it involve? If you get a sec, would love for you to walk me through it ☺️

PR #356, opened 2017-08-24
feat(cf-component-button): add 'dangerOutline' button type

hturan:
This PR adds a new `dangerOutline` button type, useful for low-contrast warning actions.

outline

### DIFF
--- a/packages/cf-component-button/src/Button.js
+++ b/packages/cf-component-button/src/Button.js
@@ -75,18 +75,20 @@ const marginTop = (marginTop, group, spaced, direction) => {
 
 const border = (theme, group, type) => {
   const isOutlineButton = type === 'dangerOutline';
-  return isOutlineButton ? {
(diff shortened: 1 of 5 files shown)


manatarms:
Just my two cents --
Is it possible to make this a more generic button? Seems like this might be a pattern. We have the type of button already, what if we just have an outline/inverse prop that just inverses the colors of the type of button?

hturan:
> Is it possible to make this a more generic button? Seems like this might be a pattern. We have the type of button already, what if we just have an outline/inverse prop that just inverses the colors of the type of button?

Totally! I originally planned to take this approach, but we're pretty sure that this will only be used for low-contrast warnings (at least for now). Also, the colors aren't inverted, we use a different (`cherry`) color for the hover-state border.

I think if we go for the `outline` prop on button in the future, we should rethink a few other things like use on darker backgrounds and our focused states 👍

PR #360, opened 2017-08-30
feat(cf-component-toast): Initial commit of basic toast

jculvey:
Creates a basic toast component for displaying inline success and error messages.

In the future, perhaps this can be enhanced to be dismissible, and anchor to different parts of the screen. For now, its a humble colored box.


jwineman:
Can someone take me off default reviewers please? :)

packages/cf-component-toast/example/basic/component.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import Toast from 'cf-component-toast';
+
+class ToastComponent extends React.Component {
+  render() {
+    return (
+      <div>
+        <Toast type="success">
```
jwineman: image

packages/cf-component-toast/src/index.js
```diff
@@ -0,0 +1,8 @@
+import { applyTheme } from 'cf-style-container';
+
+import ToastUnstyled from './Toast';
+import ToastTheme from './ToastTheme';
+
+const Toast = applyTheme(ToastUnstyled, ToastTheme);
+
+export { Toast as default, ToastUnstyled, ToastTheme };
```
tajo: please change this to:

```js
export { Toast, ToastUnstyled, ToastTheme };
```

We want to be consistent across the library with migrated components.

packages/cf-style-container/test/__snapshots__/index.js.snap
```diff
@@ -147,6 +147,8 @@ Object {
     },
     "color": Object {
       "apple": "#BD2527",
+      "apple10": "#f8e9e9",
```
manatarms: Sorry I missed this PR! 
I like this idea. What do you think of adding a comment about how the naming convention works. I'm assuming `apple10` is like saying `optacity(apple,0.1)`.

neynah: Yeah that's exactly what it means. +1 to adding a note about the naming.

PR #364, opened 2017-08-30
fix(cf-component-select): fix props warnings

vasturiano:


tajo:
According to https://github.com/JedWatson/react-select the type of value should be any.

vasturiano:
@tajo well spotted. Changed to any.

PR #366, opened 2017-09-04
fix(cf-component-card): specify react prop Title as node 

sejoker:
I have a use case when Card Title is an expander:

screen shot 2017-09-04 at 12 23 45 pm

Using prop type node fixes a warning: "Warning: Failed prop type: Invalid prop `children` of type `array` supplied to `Title`, expected `string`."

### DIFF
--- a/packages/cf-component-card/src/CardTitle.js
+++ b/packages/cf-component-card/src/CardTitle.js
@@ -12,7 +12,7 @@ const Title = ({ children, className }) => (
 );
 
 Title.propTypes = {
-  className: PropTypes.string,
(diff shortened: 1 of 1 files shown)


manatarms:
Why use <CardTitle> at all ,shouldn't it just be a link.

hturan:
Yeah, this doesn't look like a `CardTitle` to me, it looks like a `Link`. In my mind, if we want a larger link size, let's update the props on `Link`.

sejoker:
Ok, let's leave Title to be string only.

PR #367, opened 2017-09-06
Convert Icons to fela

manatarms:
I still need to do some things here. But I would appreciate a look through the code.
**BREAKING CHANGE**
 - Removes unused features spin, border, muted and white
 - Adds colors - primary,danger,warning,success,default,black,white
 - Changes font icons to svg icons
 - Removed unused type and size aliases

**TODO:**
1) Waiting for the new loading icon and close icon
2) Waiting for minified svgs (without titles, and inline styles)

This breaks some packages which I will fix. (Or feel free to update your package and fix too 🥇 )

**BREAKS:**
         _cf-ui_
```
packages/cf-builder-pagination/src/PaginationBuilder.js
packages/cf-component-heading/src/index.js
packages/cf-component-loading/src/Loading.js
packages/cf-component-pagination/example/basic/component.js
packages/cf-component-pagination/src/PaginationItem.js
```

_cf_ux_

```
packages/cf-component-navigation/src/Nav.js
packages/cf-component-search/src/Search.js
```

screen shot 2017-09-10 at 8 28 06 pm

And here's new loading gif. Sorry about the choppy framerate and my over excited mouse movement.
screenflow


manatarms:
I'd like to get this merged by tomorrow if possible. So would appreciate the reviews. I'll make PRs to fix the things that this PR breaks in cf-ui and cf-ux. But I don't want to do it here and convolute this PR.

manatarms:
cc @neynah

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,24 +1,102 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import * as svgs from './svgs.js';
+
+import { createComponent } from 'cf-style-container';
+
+const DEFAULT_HEIGHT = 15.5;
```
tajo: why 15.5?

tajo:
Well, this is not going to pass tests since Jest has no clue what `.svg` is.

```
 Cannot find module './svgs/api.svg' from 'svgs.js'
```

Also, usage of this component will require having webpack with svg loader set up. Could we turn those svg files into React components? That would make less assumptions about the consumer. For example:

carret-up.svg

```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><title>-</title><path d="M11.547,9.615l-3.3-3.3a.263.263,0,0,0-.371,0l-3.3,3.3a.263.263,0,0,0,.186.448h6.6A.263.263,0,0,0,11.547,9.615Z"/></svg>
```
into carret-up.js

```js
import React from 'react';

const CarretUp = () => <svg viewBox="0 0 16 16">
  <title>-</title>
  <path d="M11.547,9.615l-3.3-3.3a.263.263,0,0,0-.371,0l-3.3,3.3a.263.263,0,0,0,.186.448h6.6A.263.263,0,0,0,11.547,9.615Z"/>
</svg>

export default CarretUp;
```

It would also cool to export those new React component SVGs so we could directly use them in cf-component-navigation (Icon wrapper might be too restrictive for that).

packages/cf-component-icon/src/Icon.js
manatarms: That's what font-size converts them to. So this is `font-size:1rem`, 1.5rem etc etc.
If you have any better way of converting font-size:some rem to a height. I'm super open to suggestions :)

manatarms:
@tajo The idea was to keep this as maintainable as possible so design and other teams could add svgs. Let me talk with design and see if they're cool with it. Else I might need to figure a way to get these to pass.

packages/cf-component-icon/src/svgs.js
```diff
@@ -0,0 +1,99 @@
+import api from './svgs/api.svg';
+import bolt from './svgs/bolt.svg';
+import calendar from './svgs/calendar.svg';
+import caretdown from './svgs/caret-down.svg';
+import caretleft from './svgs/caret-left.svg';
+import caretright from './svgs/caret-right.svg';
+import caretup from './svgs/caret-up.svg';
+import chart from './svgs/chart.svg';
+import clipboard from './svgs/clipboard.svg';
+//TODO add actual close svg [Used once on notifications]
```
hturan: Can we close out these `TODO`s?

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,24 +1,102 @@
...
+  }),
+  'span',
+  ['aria-label']
+);
+
 class Icon extends React.Component {
   render() {
-    let className = `cf-icon cf-icon--${this.props.type}`;
+    const { className, size, label, color } = this.props;
+    const icon = svgs[this.props.type.replace(/-/g, '')];
```
hturan: Not sure how I feel about this. Can't we just have our icon types match our icon names?

packages/cf-component-icon/src/Icon.js
```diff
@@ -1,24 +1,102 @@
...
+    let height, fill;
+
+    if (!icon) throw `Invalid Icon type ${this.props.type}`;
```
hturan: Do we need this? Surely `PropTypes` will warn against this too.

packages/cf-component-icon/src/svgs/api.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><title>-</title><path d="M15.124,12.749v-9.5A1.826,1.826,0,0,0,13.343,1.47H2.657A1.826,1.826,0,0,0,.876,3.251v9.5A1.826,1.826,0,0,0,2.657,14.53H13.343A1.826,1.826,0,0,0,15.124,12.749Zm-9.26-2.493L5.507,9.069H3.667L3.31,10.256H2.123L3.9,5.151H5.21l1.781,5.105ZM5.27,8.178c-.356-1.068-.534-1.721-.594-1.84s-.059-.3-.119-.356c-.059.3-.3,1.009-.653,2.2H5.27Zm5.936-1.425A1.534,1.534,0,0,1,10.671,8a2.21,2.21,0,0,1-1.484.415H8.713V10.2H7.644V5.21h1.6a2.145,2.145,0,0,1,1.425.415,1.378,1.378,0,0,1,.534,1.128Zm-2.434.831h.356A1.346,1.346,0,0,0,9.9,7.407a.626.626,0,0,0,.238-.594.725.725,0,0,0-.178-.534A.955.955,0,0,0,9.306,6.1H8.831V7.585h-.06Zm3.384,2.671V5.21h1.068v5.105H12.155Z"/></svg>
```
hturan: What's with the `<title>-</title>` tags?

hturan:
> It would also cool to export those new React component SVGs so we could directly use them in cf-component-navigation (Icon wrapper might be too restrictive for that).

If we can't use our `<Icon>` component for a specific purpose, then we either need to change how it's being used or how it works.

hturan:
Ace work here, @manatarms! 🎉

packages/cf-component-icon/src/svgs.js
manatarms: Yup. On it :)

packages/cf-component-icon/src/Icon.js
manatarms: I want to do that. Instead of doing a bunch of things to check for a match.

packages/cf-component-icon/src/Icon.js
manatarms: True. I put this in for debugging.

packages/cf-component-icon/src/svgs/api.svg
manatarms: New icons have them removed.

tajo:
@hturan `cf-component-logo` assumes url-loader loader. 

>  it seems a core requirement here is that we maintain raw SVG files

Why?

manatarms:
One more look please?

manatarms:
Arggg!!! I guess I need to upgrade all the cf-ui components that need icon as well :(

manatarms:
@hturan could you take another look. I think I've addressed all of your comments.

manatarms:
I will also do loading as part of this PR. So I have an easier time migrating it simultaneously in www-next.

packages/cf-component-pagination/src/PaginationItem.js
```diff
@@ -92,7 +92,8 @@ class PaginationItem extends React.Component {
     if (isEllipsis) {
       children = <span>…</span>;
     } else if (isLoading) {
-      children = <Icon type="loading" label={false} />;
+      // TODO Change this to new Loading component
```
manatarms: Reminder to self. Do this before merging.

packages/cf-component-loading/src/Loading.js
```diff
@@ -1,8 +1,45 @@
 import React from 'react';
-import Icon from 'cf-component-icon';
+import { Icon } from 'cf-component-icon';
+
+import { createComponent } from 'cf-style-container';
+
+const LoadingWrapper = createComponent(({}) => ({
+  width: 40,
```
hturan: This seems wider than our previous default. Have you tried implementing this yet and seeing how it looks?

packages/cf-component-loading/src/Loading.js
```diff
@@ -1,8 +1,45 @@
...
+  animationDuration: '2s',
+  animationIterationCount: 'infinite',
+  animationTimingFunction: 'ease',
+  '>svg': {
+    height: '100%',
+    fill: 'transparent',
+    stroke: '#ff7900',
```
hturan: Should this come from the theme?

hturan: I'm worried about our default `<Loading />` component (that will be used everywhere) being a 40px-wide orange circle. How does this look inside an orange `<Button />`? Have we audited where we currently use the spinner and made sure that this color works?

packages/cf-component-loading/src/Loading.js
manatarms: Aiye. My bad. Forgot I had this left to do. So Nena is going to give me some new animation values. Which I will swap out. But the loading icon will have 
3 colors - default,white, orange
All sizes allowed by cf-component-icon
3 types -  icon, loading text and opacity.

packages/cf-component-loading/src/Loading.js
manatarms: Yup :)

hturan:
@manatarms I'm super happy with the `<Icon />` changes! My only concern now lies around the new `<Loading />` component, outlined in the comment above.

manatarms:
@hturan Everything is done in this. Could you take a look and merge this for me if you approve. I don't want to deal with any merge conflicts tomorrow :)

packages/cf-component-loading/src/Loading.js
```diff
@@ -1,8 +1,49 @@
...
+  return {
+    width: 15,
+    animationDuration: '2.5s',
+    animationIterationCount: 'infinite',
+    animationTimingFunction: 'linear',
+    '>svg': {
+      height: '100%',
+      fill: 'transparent',
+      strokeWidth: 6,
+      stroke: theme.colors.hail,
```
hturan: You mentioned before that this will have multiple colour options. Did we scrap that idea?

neynah: I figured we should put more thought into where folks might want to use the loading spinner before we propose other color/style options. We still want to add multiple options in the near future.

PR #369, opened 2017-09-07
feat(cf-component-select): update styles

marksteyn:
Move all styles to fela.

Prevent page scrolling the scrolling the menu.

### DIFF
--- a/packages/cf-component-select/src/MenuHeader.js
+++ b/packages/cf-component-select/src/MenuHeader.js
@@ -5,12 +5,12 @@ import { createComponent } from 'cf-style-container';
 
 const MenuHeader = createComponent(
   ({ theme }) => ({
-    color: '#454545',
+    color: theme.color.dusk,
     cursor: 'default',
     fontWeight: 'bold',
     fontSize: '.86667rem',
-    backgroundColor: '#f9f9f9',
-    borderBottom: `1px solid #eeeeef`,
+    backgroundColor: theme.color.moonshine,
+    borderBottom: `1px solid ${theme.color.dust}`,
     width: '100%',
     '> *': {
       display: 'inline-block',
--- a/packages/cf-component-select/src/MenuHeader.js
+++ b/packages/cf-component-select/src/MenuHeader.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactSelect from 'react-select';
+import { createComponent } from 'cf-style-container';
+
+const MenuHeader = createComponent(
+  ({ theme }) => ({
+    color: '#454545',
+    cursor: 'default',
+    fontWeight: 'bold',
+    fontSize: '.86667rem',
+    backgroundColor: '#f9f9f9',
+    borderBottom: `1px solid #eeeeef`,
+    width: '100%',
+    '> *': {
+      display: 'inline-block',
+      padding: '8px 10px'
+    }
+  }),
+  'h3'
+);
+
+export default MenuHeader;
--- a/packages/cf-component-select/src/ScrollContainer.js
+++ b/packages/cf-component-select/src/ScrollContainer.js
@@ -41,9 +41,9 @@ class ScrollContainer extends React.Component {
   }
 }
 
-const ScrollContainerWrapper = createComponent(() => {
+const ScrollContainerWrapper = createComponent((theme) => {
   return {
-    maxHeight: '198px',
+    maxHeight: theme.maxMenuHeight || '198px',
     overflowY: 'auto'
   };
 }, ScrollContainer);
--- a/packages/cf-component-select/src/ScrollContainer.js
+++ b/packages/cf-component-select/src/ScrollContainer.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { createComponent } from 'cf-style-container';
+
+class ScrollContainer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleMouseOver = this.handleMouseOver.bind(this);
+    this.handleMouseOut = this.handleMouseOut.bind(this);
+  }
+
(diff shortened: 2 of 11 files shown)


packages/cf-component-select/src/MenuHeader.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactSelect from 'react-select';
+import { createComponent } from 'cf-style-container';
+
+const MenuHeader = createComponent(
+  ({ theme }) => ({
+    color: '#454545',
```
tajo: should be a constant (on of `theme.colors`)

tajo: (all colors should be constants)

marksteyn: How are new colors named?

packages/cf-component-select/src/Select.js
```diff
@@ -43,7 +105,7 @@ class Select extends React.Component {
     }
 
     return (
-      <div className={className}>
+      <div className={`cf-select ${className}`}>
```
sejoker: is it due interoperability you adde `cf-select`. I thought  we could migrate those styles into fela as well

packages/cf-component-select/src/ScrollContainer.js
```diff
@@ -0,0 +1,51 @@
...
+      >
+        {this.props.children}
+      </div>
+    );
+  }
+}
+
+const ScrollContainerWrapper = createComponent(() => {
+  return {
+    maxHeight: '198px',
```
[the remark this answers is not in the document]
marksteyn: All the styles come from the styles.css

sejoker:
Test are passing, Linting is not.

packages/cf-component-select/src/Select.js
```diff
@@ -1,15 +1,58 @@
+/**
+ * original html structure (note that extra divs have been inserted by using fela)
```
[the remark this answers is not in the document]
marksteyn: A lot of the styles are still dependent on the html structure using '& child' rules.

packages/cf-component-select/src/Select.js
marksteyn: There are components in cf-www-next which assume the existence cf-select.

sejoker: I see 2 such instances, e.g. 
```
  '> .cf-select': {
    flex: 1,
    margin: '0 5px'
  },
```

 we should allow to configure this  through Theming.

```
import { SelectUnstyled, SelectTheme } from 'cf-component-select';

const CustomSelect = applyTheme(SelectUnstyled, SelectTheme, { margin: '0 5px',  flex: 1 });
```

manatarms:
> Prevent page scrolling the scrolling the menu.
OMG thank you.

packages/cf-component-select/src/Select.js
```diff
@@ -76,7 +148,143 @@ Select.propTypes = {
...
+    display: theme.display,
+    verticalAlign: theme.verticalAlign,
+    position: 'relative',
+    opacity: disabled ? 0.5 : 1,
+
+    '& .Select-control': {
+      backgroundColor: disabled ? theme.colorGrayDark : theme.colorWhite,
+      border: `1px solid ${theme.color.hail}`,
+      borderRadius: theme.borderRadius,
+      color: theme.colorGrayDark,
```
manatarms: We're avoiding using the colors outside the colors object. https://cloudflare.github.io/cf-ui/#cf-design-colors should be the only colors you need. If you need a new color we can add it.

packages/cf-component-select/src/Select.js
```diff
@@ -76,7 +148,143 @@ Select.propTypes = {
...
+      transition: 'border-color 0.2s ease'
+    },
+    '& .Select-control .Select-input:focus': {
+      outline: 'none'
+    },
+    '& .Select-menu-outer': {
+      position: 'relative'
+    },
+    '&:hover .Select-control': {
+      borderColor: theme.colorBlueDark
```
manatarms: I'm going to add the new hover colors today. So we can swap these out.

packages/cf-component-select/src/Select.js
```diff
@@ -76,7 +148,143 @@ Select.propTypes = {
...
+    },
+    '& .is-focused .Select-control': {
+      borderColor: theme.colorBlueDark
```
manatarms: Same

manatarms:
Looks good! Thanks for this.

packages/cf-component-select/src/Select.js
marksteyn: Surely we should remove anything from the theme that shouldn't be used, or at least mark it deprecated?

manatarms: Yeah. I'm working on doing that. Been busy with migration things. But I'll get to our color stuff soon.

marksteyn: Why were the added to theme to start with?

manatarms: I think they were added before we had the new styleguide stuff in place. I think this will change once the new new styleguide lands.

manatarms:
I didn't see a `BREAKING CHANGE` in any of the commit messages. We should do a major version bump for this.

marksteyn:
This is my first time using conventional comments - do we have any documentation in the wiki for it?

packages/cf-component-select/src/Select.js
marksteyn: Can you add black and white as well?

manatarms:
This should help https://conventionalcommits.org/
But I agree with you. We should add some docs about it. And also enforce it on CI :(
This is how I do breaking changes https://github.com/cloudflare/cf-ui/pull/367/commits/e954159b60c3656b5e386373303c870b79635850

PR #370, opened 2017-09-11
fix(cf-component-select): Bump version. 

marksteyn:
BREAKING CHANGE: because it didn't update the version correctly last time


hturan:
Try not to break too much with that comment change 😅

PR #373, opened 2017-09-14
So many proptype warnings

manatarms:
Fixed propType warnings. Console is now clean.


tajo:
I was dreamin' about this for weeks. 👏

jculvey:
Looks good

PR #376, opened 2017-09-15
feat(cf-style-container): add applyStaticStyles method

vasturiano:
A HOC that applies a string of static styles to a component using fela's renderStatic. Useful for integration with older libraries that require side loading of a static CSS block.

### DIFF
--- a/packages/cf-style-container/README.md
+++ b/packages/cf-style-container/README.md
@@ -145,6 +145,20 @@ const MyComponent = ({ theme }) => {
 export default withRenderer(MyComponent);
 ```
 
+### applyStaticStyles(staticStyles, Component)
+
+A HOC that applies a string of static styles to a component using fela's [renderStatic](https://github.com/rofrischmann/fela/blob/master/docs/api/fela/Renderer.md#renderstaticstyle-selector).
+Useful for integration with older libraries that require side loading of a static CSS block.
+
+```jsx
+import { applyStaticStyles } from 'cf-style-container';
+
+const staticStyles = '.purple-component { background-color: purple }';
+const MyComponent = () => <div className='purple-component' />;
+
+export default applyStaticStyles(staticStyles, MyComponent);
+```
+
 ## mergeTheme(baseTheme, ...themes)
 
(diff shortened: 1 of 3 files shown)


hturan:
Heya!

Could we get some further context for this change?

- The `fela` [documentation](https://github.com/rofrischmann/fela/blob/master/docs/api/fela/Renderer.md#renderstaticstyle-selector) isn't super clear about what `renderStaticStyle` actually _does_. Is it really just adding a static class name to the renderer's `<style>` node? How do we avoid leaky CSS abstractions with this approach (e.g. in the example, if I go and add a `.purple-component` div elsewhere in the app, does it inherit those styles)?

- When do we use this approach, and when do we spend two weeks dissecting `cf-component-select` and re-writing every part of it from the ground-up in `fela`? Have we got examples where one is too much work?

- Should we finish off the `fela` migration by just taking the `cf-component-table` & `cf-component-modal` CSS and pasting it into a component like this?

- This is now the fifth way we sanction styling components in `cf-ui`: `createComponent`, `withRenderer`, `applyTheme`, `cf-component-box` and now  `applyStaticStyles`.

The documentation for this package is 😍 lovely 😍 and descriptive, but I'm still worried about the multitude of options available.

tajo:
**This is the last resort option** for supporting 3rd party libraries that export a stylesheet and there is no other way how to style them (typically some viz libraries). Also note, that this is only a convenience wrapper around `withRenderer` so we are not enabling a new pattern.

I am not very happy about using libraries like that and I'm trying to always avoid them but not every library has a better API (allowing to pass down own classNames). 

> Should we finish off the fela migration by just taking the cf-component-table & cf-component-modal CSS and pasting it into a component like this?

No. That would be cheating. :)

> When do we use this approach, and when do we spend two weeks dissecting cf-component-select and re-writing every part of it from the ground-up in fela? Have we got examples where one is too much work?

We use CSS in JS for some good reasons. I believe they still hold. `renderStatic` is not really CSS in JS, just a different way how to load plain CSS stylesheet.

